PR #68, opened 2023-09-23
C deploy to flyio

catflydotio:


examples/fly/Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN --mount=type=cache,target=/usr/local/cargo,from=rust:latest,source=/usr/loca
 # Runtime image
 FROM debian:bookworm-slim
 
-RUN apt update && apt install -y sqlite3 && rm -rf /var/lib/apt/lists/*
+RUN apt update && apt install -y sqlite3 watch && rm -rf /var/lib/apt/lists/*
```
jeromegn, line 30: I assume this is for the `watch` part of the quickstart.

I'd advise running `watch` locally and using `fly proxy 8080` to reach the API of the Fly-hosted Corrosion! You can then use `corrosion template --api-addr 127.0.0.1:8080 ...`

catflydotio: Yes, it's for `corrosion template`. :) I've been writing that example up with two local terminals, each running `fly ssh console` on one of two Machines. 

Can you expand on that recommendation? 

I do agree that we should illustrate `fly proxy` for connecting a local machine to Fly-hosted Corrosion, but 
having it all self-contained on Fly.io appeals to me in a first demo -- you're pulling levers and observing propagation directly on the hosts in the cluster. And you don't have to install Corrosion locally. 

But you may be able to convince me otherwise.

Are you concerned that I'm loading up the Dockerfile with stuff we would only use for a demo? We could also have a Dockerfile for learning and a separate Dockerfile (without watch and sqlite3 and the example schema, etc.) that you'd use as a clean-ish base for deploying an app with Corrosion for real.

catflydotio:
Preview: https://c-corrodocs.fly.dev/deploy-on-fly/index.html

catflydotio:
This is only markdown and example files so I'm betting safe to merge.

Issue #106, opened 2023-12-02
Streaming rows as comma-separated JSON that form and array


jeromegn:
Not all languages / frameworks / libraries have an easy time dealing with NDJSON out of the box, even if they have streaming capabilities.

It might be interesting to explore and alternate format: streaming as a normal JSON array.

How? First we send the `[` character, then we send each row as an independent JSON object (well, an array? could be configurable to set the column names properly), send a comma in between, send a `]` at the end.

This way we don't have to hold onto all the rows in-memory but can still stream a normal JSON array.

jeromegn:
I think this might be more practical if there was a way to specify more settings about the expected format, e.g. "just send me JSON objects in an array" where the column names would be used as fields in the JSON object, no need for a `columns` object at the start.

`eoq` is not be necessary, but it doesn't hurt anything. Possibly something configurable too.

The reason for NDJSON is so clients don't need to buffer large quantities of data. It's not always possible or required to do that so the alternative comma-separated format sounds nice.

I think the default format for a client that just wants a big valid JSON array should be:

```json
[
  {"col1": "value", "col2": 12345, "col3": null},
  {"col1": "value", "col2": 12345, "col3": null},
  {"col1": "value", "col2": 12345, "col3": null}
]
```

because they'll probably have to re-arrange the data themselves if we're sending it in a different format.

gorbak25:
NDJSON is the standard streaming format and is trivially parseable in any language (split on newline, parse each line). Client support in 2026 improved a lot.

PR #110, opened 2023-12-04
basic implementation of comma-separated json

lillianberryfly:
#106

### DIFF
--- a/crates/corro-agent/src/api/public/mod.rs
+++ b/crates/corro-agent/src/api/public/mod.rs
@@ -16,6 +16,7 @@ use hyper::StatusCode;
 use itertools::Itertools;
 use metrics::counter;
 use rusqlite::{named_params, params_from_iter, ToSql, Transaction};
+use serde::Deserialize;
 use spawn::spawn_counted;
 use tokio::{
     sync::{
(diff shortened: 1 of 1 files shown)


crates/corro-agent/src/api/public/mod.rs
```diff
@@ -474,6 +481,13 @@ pub async fn api_v1_queries(
     let (data_tx, mut data_rx) = channel(512);
 
     tokio::spawn(async move {
+        if !pms.ndjson {
```
jeromegn, line 484: I think we could do `Accept: application/ndjson` as a header. At least it's generally how content type is negotiated :)

I just invented that, but maybe there's a better mime type for that.

crates/corro-agent/src/api/public/mod.rs
```diff
@@ -474,6 +481,13 @@ pub async fn api_v1_queries(
...
+            if let Err(e) = tx.send_data(bytes::Bytes::from_static(b"[")).await {
```
jeromegn, line 485: It would be cleaner to implement a `tokio_util::codec::{Encoder,Decoder}` for this format!

I wouldn't hold back this PR for that though!

PR #117, opened 2023-12-07
Add external_addr to gossip configuration (fixes #115)

### LINKED ISSUE #115: Add a setting for `external_addr` in the `gossip` config
Some network setups might need to listen on a different address than the one Corrosion nodes will be connecting to. Adding a setting for an optional `external_addr` (defaults to the value of `addr`) in the `gossip` config could fix that.

For example: The problem happens if the `addr` is set as the completely valid value of `[::]:<port>` and other nodes connect to an IP they can reach, e.g. `[::1]:<port>`.

We had not encountered this issue before because our `addr` was set to an address available for local binding and also available externally (a wireguard interface, in this case). However, it is common for users to set `[::]` as an IP if they don't want to bother figuring out which one they should bind to.

spacekookie:


crates/corro-agent/src/api/peer.rs
```diff
@@ -215,7 +215,10 @@ async fn build_quinn_server_config(config: &GossipConfig) -> eyre::Result<quinn:
 pub async fn gossip_server_endpoint(config: &GossipConfig) -> eyre::Result<quinn::Endpoint> {
     let server_config = build_quinn_server_config(config).await?;
 
-    Ok(quinn::Endpoint::server(server_config, config.bind_addr)?)
+    Ok(quinn::Endpoint::server(
+        server_config,
+        config.external_addr.unwrap_or(config.bind_addr),
```
jeromegn, line 220: I think we always want to bind to `config.bind_addr`!

That logic should be used here instead (maybe with a new `Agent` function for it): https://github.com/superfly/corrosion/blob/6d1b5ecc890a16963209b103b605f0d172f4d52c/crates/corro-agent/src/agent.rs#L442

PR #122, opened 2023-12-13
Refactor agent

spacekookie:
This commit breaks apart the giant `agent.rs` file into a set of smaller (albeit still big) modules. Hopefully this makes the code easier to reason about and work on for anyone newly coming into the project (like me :grin:)


crates/corro-agent/src/agent/handlers.rs
```diff
@@ -0,0 +1,505 @@
...
+                start.elapsed().as_secs_f64()
+            );
+        }
+        Ok::<_, eyre::Report>(())
+    })?;
+    debug!("done handling db_cleanup");
+    Ok(())
+}
+
+// DOCME: provide function context
```
jeromegn, line 359: This is the function that bundles changes to optimize transactions sizes with SQLite which makes everything more efficient at the cost of a little bit of waiting for more data before processing it.

It's used by both broadcast receivers and sync receivers so it's all one happy family.

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -0,0 +1,505 @@
...
+        }
+    }
+
+    // process the last changes we got!
+    if let Err(e) = util::process_multiple_changes(&agent, buf).await {
+        error!("could not process multiple changes: {e}");
+    }
+}
+
+// DOCME: add context
```
jeromegn, line 430: Handles **starting** a sync with multiple other nodes at once. It uses a simple algorithm to choose "good" candidates based on RTT and how many versions we need from a specific node. Sprinkle in a little randomness so we don't keep syncing with the same ones.

crates/corro-agent/src/agent/run_root.rs
```diff
@@ -0,0 +1,241 @@
+//! Start the root agent tasks
+//!
+//! 1. foo
+//! 2. bar
+//! 3. spawn `super::other_module`
```
jeromegn, line 5: Haha, what does this mean?

crates/corro-agent/src/agent/uni.rs
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+    });
+}
+
+/// Start an async buffer task that pull messages from one channel
+/// (process_uni_rx) and moves them into the next if the variant
+/// matches (bcast_msg_tx)
+// DOCME: why is this task needed?  Do we just want to get this
+// function out of the stream hot path?
```
jeromegn, line 90: It is possibly not needed. I think it used to do more, but not so much anymore!

However, we definitely don't want to slow down receiving broadcasts in any way. We could spawn tasks for each broadcast we get, this is likely more conservative with resources.

crates/corro-agent/src/agent/util.rs
```diff
@@ -0,0 +1,1479 @@
...
+    };
+
+    if !passed {
+        return Err(axum::http::StatusCode::UNAUTHORIZED);
+    }
+
+    Ok(next.run(request).await)
+}
+
+// DOCME: provide some context for this function
```
jeromegn, line 404: This is a fun one.

When we insert `crsql_changes` we get via broadcasts and sync process, we don't know which changes from the past might've been invalidated (cleared). Sometimes whole versions are cleared and we don't know until we make these expensive queries (they're indexed at least).

When we don't know about cleared versions, we accumulate much more data in the `__corro_bookkeeping` table and, more importantly, the structure in-memory for "current" versions (not cleared) is a lot bigger. Cleared versions are stored both in the DB and in-memory as ranges like "versions `1..=5` for actor-id `x` don't appear in the DB anymore" and therefore only use minimal space.

If we don't find these cleared versions periodically, memory keeps growing as more changes are processed (especially bad for many small changes being processed). Finding them in the DB lets us rewrite the bookkeeping DB in a more compact format and the same goes for the in-memory DB.

crates/corro-agent/src/agent/util.rs
```diff
@@ -0,0 +1,1479 @@
...
+
+    let cleared_db_versions: BTreeSet<CrsqlDbVersion> = tx
+        .prepare_cached(&to_clear_query)?
+        .query_map(params_from_iter(params.into_iter()), |row| row.get(0))?
+        .collect::<rusqlite::Result<_>>()?;
+
+    Ok(cleared_db_versions)
+}
+
+// DOCME: provide some context for this function
```
jeromegn, line 458: This is a loop that handles periodically initiating a sync from the current node to many other nodes (via another function), but also applying "buffered" (partial) changesets (like partial transactions).

It works this way so we try and process these buffered changes before we try to sync so we have less to sync.

crates/corro-agent/src/agent/util.rs
```diff
@@ -0,0 +1,1479 @@
...
+                    Err(e) => {
+                        error!(%actor_id, %version, "could not apply fully buffered changes: {e}");
+                    }
+                }
+            }
+        }
+    }
+}
+
+// DOCME: provide some context for this function
```
jeromegn, line 538: Compacts the database by finding cleared versions. One actor at a time to lock for a minimal amount of time.

crates/corro-agent/src/agent/util.rs
```diff
@@ -0,0 +1,1479 @@
...
+
+                tokio::time::sleep(Duration::from_secs(2)).await;
+            }
+
+            Ok::<_, eyre::Report>(())
+        });
+    }
+}
+
+// DOCME: provide some context for this function
```
jeromegn, line 593: When we receive an empty changeset or when we find them by periodically calling the `find_cleared_db_versions` function, we immediately apply them to our in-memory state, but we don't do it in the DB. This is to reduce DB load and be more efficient by chunking these operations.

jeromegn (commented):
Left some notes, haven't yet looked to make sure everything should work. I'm pretty sure it will if the tests pass! I'll give it another go next week.

crates/corro-agent/src/agent/run_root.rs
spacekookie: Uuuuh :see_no_evil: I think I added that comment very early into the refactor and then kinda forgot about it :joy:

jeromegn:
I'm still seeing a significant issue with "sending sync": 
image

I'm fairly sure it's related to the changes in this PR and not an artifact of deploying. Switching back to `main` fixes it.

jeromegn:
> I'm still seeing a significant issue with "sending sync"

I've seen this happen w/ a different branch (crsqlite 0.16) and so I'm inclined to say it doesn't have anything to do with your branch.

I've also seen a node recover from this "lower sent sync" situation and it happened exactly as the node "rejoined" the cluster. That's aligned with a theory I had where a node might've been considered down or something like that, preventing others from syncing with it and therefore that node would not be sending as many changes because none were requested.

I'm not sure how this happens to be honest! Something to investigate...

jeromegn:
@spacekookie I have merged the crsqlite 0.16 branch and you'll have to figure out how to put the changes in the right spots. I apologize!

jeromegn:
I lied, I ended up handling the merge conflicts. It wasn't too bad!

jeromegn:
I'm pretty sure this is working fine and I don't want to complicate development by having to keep merging with main. I'm merging this and we can improve the docs later!

PR #125, opened 2023-12-19
doc: add section on setting up a development environment

spacekookie:


doc/devel.md
```diff
@@ -0,0 +1,24 @@
+# Setup development environment
+
+This section contains some information that might be relevant for you when setting up a `corrosion2` development environment.  Depending on your computer setup your experience will vary.
```
jeromegn, line 3: We should refer to Corrosion as just "corrosion". V2 is kind of internal to differentiate with the "proof of concept" that we had running before.

jeromegn (changes requested):

doc/SUMMARY.md
```diff
@@ -44,4 +45,4 @@
     - [api]() (to come)
     - [admin]() (to come)
     - [telemetry]() (to come)
-    - [consul]() (to come)
\ No newline at end of file
+    - [consul]() (to come)
```
spacekookie, line 48: I'm not entirely sure I understand what this diff is but I'll try to remove it in a bit

jeromegn (approved):

PR #128, opened 2023-12-28
cr-sqlite 0.16

jeromegn:
Upgrades and migrates to cr-sqlite 0.16 which mostly includes the same features we had been running before, with a few differences:
- `site_id` is not nullable anymore on `*__crsql_clock` tables.
- a bit more of it is now written in Rust

I have used a release from `main` which includes a feature I was able to [get merged in](https://github.com/vlcn-io/cr-sqlite/pull/415). Essentially: every change needs a winner, the ultimate tie-breaker is the `site_id` in the event that both changes are identical in every other way.


jeromegn:
Has been running on a significant portion of the fleet for a few days now. Merging.

PR #138, opened 2024-01-10
api: disable Nagle's algorithm for HTTP API

pborzenkov:
This adds a rather noticeable latency ~40ms to each HTTP request if
there is a constant stream of them.


soulware:
Related: https://www.youtube.com/watch?v=mqvVmYhclAg
> surely one of these should not be the default in 2023

jeromegn (approved):

PR #142, opened 2024-01-26
Additional metrics to debug broadcast lag

spacekookie:


crates/corro-types/src/channel.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl<T: Debug> Debug for CorroReceiver<T> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        self.inner.fmt(f)
+    }
+}
+
+/// Create a bounded channel which tracks occupancy with a label
+pub fn bounded<T>(max: usize, label: &'static str) -> (CorroSender<T>, CorroReceiver<T>) {
+    let gauge = gauge!("corro.runtime.channel_occupancy", "max_capacity" => max.to_string(), "label" => label);
```
jeromegn, line 47: Label `label` called "label" eh? I assume that's fine.

What's the `max_capacity` for? If it's for charting, it might be more efficient to use a separate metric like `corro.runtime.channel.max_capacity` and set it once when initializing.

Do we care for tracking the "occupancy" vs how much is left? If it was instead called `channel.capacity`, we could more easily alert when it reaches zero.

crates/corro-types/src/channel.rs
```diff
@@ -0,0 +1,107 @@
...
+            gauge: gauge.clone(),
+            inner: tx,
+        },
+        CorroReceiver { gauge, inner: rx },
+    )
+}
+
+impl<T> CorroSender<T> {
+    pub async fn send(&self, value: T) -> Result<(), SendError<T>> {
+        self.inner.send(value).await.map(|r| {
```
jeromegn, line 61: We might as well track how long it took to send the value into the channel. This would tell us if there's backpressure from channels and how much. This gives us an extra bit of info compared to knowing that a channel does not have capacity anymore.

jeromegn (commented):
This looks generally good, but I left a few comments that we should discuss.

One other thing I'm wondering about: instead of performing a gauge incr/decr on every channel send/recv, we could clone the sender once and observe its capacity on a fixed interval or on metrics collection.

crates/corro-types/src/channel.rs
spacekookie: Yeah, we have some channels that use different sizes. I thought it would be useful to see how much of a channel is actually in use vs just the raw usage number. Happy to change any of the actual metric names (I'm new to this stuff anyway)

jeromegn (approved):

PR #144, opened 2024-01-29
Change propagation improvements

jeromegn:
This PR makes a few improvements:
- Uses a `VecDeque` instead of `Vec` to process needed versions when parallel syncing. It's more efficient to `pop_front` than to `Vec::drain`.
- Broadcasts more aggressively
- Fixes an issue with a wild libssl appearance in the binary's dependencies
- Removes 2 whole channels (bcast_msg and process_uni) and instead encompass the functionality elsewhere
- Will concurrently process batches of changes (max concurrency = 3), in the background. This should prevent the channel from reaching zero capacity.


crates/corro-types/Cargo.toml
```diff
@@ -49,6 +49,7 @@ tracing = { workspace = true }
 tripwire = { version = "0.1.0-alpha.0", path = "../tripwire" }
 uhlc = { workspace = true }
 uuid = { workspace = true }
+cuckoofilter = { version = "0.5.0" }
```
spacekookie, line 52: Is this still needed, considering the code using it was commented out?

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -401,45 +376,127 @@ pub async fn handle_changes(
...
+
+            // process these first, we don't care about the result,
+            // but we need to drain it to free up concurrency
+            res = join_set.join_next(), if !join_set.is_empty() => {
+                debug!("processed multiple changes concurrently");
+                if let Some(Err(e)) = res {
+                    error!("could not process multiple changes: {e}");
+                }
+                continue;
+            },
```
spacekookie, line 425: This whole section makes me wonder if we would be better off using a mpmc channel (like `crossbeam_channel`, although not sure how they play with async, otherwise there's also `async-channel` which is mpmc) here instead to concurrently read from the channel and apply changes.

Is batching that important for performance?

This is a tangent but it's frustrating that only the tokio channels implement `recv_many`. I'm wondering if there's some fundamental limitation in mpmc channel designs that don't allow receiving many messages

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -401,45 +376,127 @@ pub async fn handle_changes(
     mut rx_changes: CorroReceiver<(ChangeV1, ChangeSource)>,
     mut tripwire: Tripwire,
 ) {
+    const MIN_CHANGES_CHUNK: usize = 2000;
+    let mut queue: VecDeque<(ChangeV1, ChangeSource)> = VecDeque::new();
     let mut buf = vec![];
     let mut count = 0;
 
+    const MAX_CONCURRENT: usize = 3;
```
spacekookie, line 384: It could make sense to update this number based on the capacity of the `rx_changes` channel

spacekookie (commented):
Looks good overall. The spawn logic around applying changes is pretty complex and I'm wondering if we can find a way to get this performance with a simpler design. But it definitely works for now

crates/corro-agent/src/agent/handlers.rs
jeromegn: I did wonder about that :) I wouldn't bother just yet but I think it's a good idea.

crates/corro-agent/src/agent/handlers.rs
jeromegn: Batching inserts is important for SQLite performance yes. Beginning a transaction, inserting a single row, committing a transaction is a bit expensive. There's also some overhead is sending changes to subscriptions which each have to check if the changes might match their subscription and then apply logic.

How do you see a mpmc channel working here?

crates/corro-types/Cargo.toml
jeromegn: Ah yeah, I started on something and didn't follow through. I'll remove!

crates/corro-agent/src/agent/handlers.rs
spacekookie: Ah right, that makes sense.

With an mpmc channel I would invert the batching logic: push changes to a channel, have three tasks pull them from the channel and into a buffer, then apply in-line when there's enough changes. If memory drop pressure is an issue we could even make them dedicated threads (although that might be overkill).

It _should_ result in a similar performance profile but simplify the big tangled function which receives from the channel, then applies to a buffer, then spawn tasks with buffers as needed.
(Changes wouldn't be applied in order, but I assume the order that changes are appearing on the node isn't in-order anyway so that wouldn't matter much?)

PR #150, opened 2024-02-04
Configurable client addr for outgoing socket bind

jeromegn:
Using an unspecified IP can sometimes cause issues, especially if your hosts have a lot of network interfaces. With this configuration option, it's possible to set a specific addr or a specific ip with port 0 to keep the pooling features.

Even on hosts with only a few network interface, this can result in faster `sendmmsg` calls.


spacekookie (approved):
LGTM

PR #152, opened 2024-02-07
Implement priority broadcasts for local changes

spacekookie:
Initial implementation of priority broadcasts and cleaning up the processing_changes function pipeline a bit. Some open questions in this implementation:

**Are there/ what are better ways of exposing a "priority insert" over the API?**

My rationale was to allow certain changes to "jump the queue" for several broadcast cycles if it is deemed important enough. It would be up to the caller to decide which insert is important. This does add a lot of overhead in the API though.

**Should priority inserts still use a queue?**

Currently every priority broadcast will immediately insert the received change (in an async task). When receiving a series of priority inserts, maybe it would make sense to collect a few (less than 20) together to make inserts more efficient. 

I'll have a look at the CI failure. I had to update a few of the tests to make them pass again, because of the change to `Statement`. I would have expected the updated `From<&str>` implementation to take care of it but :shrug:


crates/corro-types/src/broadcast.rs
```diff
@@ -30,13 +30,22 @@ use crate::{
...
 pub enum UniPayloadV1 {
     Broadcast(BroadcastV1),
 }
 
+/// Broadcast change payload that supports priority
+#[derive(Debug, Clone, Readable, Writable)]
+pub enum UniPayloadV2 {
+    Broadcast(BroadcastV1),
+    PriorityBroadcast(BroadcastV1),
+}
```
jeromegn, line 47: I don't think we need a V2 here. We might even be able to keep a single variant in V1!

I made this little test to check if things could be backwards compatible:

```rust
    #[test]
    fn shenanigs() {
        #[derive(Debug, Readable, Writable)]
        enum A {
            A(Inner),
        }

        #[derive(Debug, Readable, Writable)]
        struct Inner {
            foo: String,
        }

        #[derive(Debug, Readable, Writable)]
        enum B {
            A {
                inner: Inner,
                #[speedy(default_on_eof)]
                flag: bool,
            },
        }

        let v = A::A(Inner { foo: "bar".into() }).write_to_vec().unwrap();

        let b = B::read_from_buffer(&v).unwrap();

        println!("b: {b:?}");

        let v2 = b.write_to_vec().unwrap();
        let a = A::read_from_buffer(&v2).unwrap();

        println!("a: {a:?}");
    }
```

crates/corro-types/src/broadcast.rs
```diff
@@ -80,6 +89,7 @@ pub enum BroadcastV1 {
 pub enum ChangeSource {
     Broadcast,
     Sync,
+    PriorityBroadcast,
```
jeromegn, line 92: I don't think we ever match on this, so I'd be in favor of not having it.

jeromegn (changes requested):

crates/corro-types/src/broadcast.rs
spacekookie: Oh yeah, good catch. I thought I had already removed it

spacekookie:
I'll rebase this again once #134 is merged, just wanted to get the conflicts out the way

jeromegn (approved):
Looking good!

PR #156, opened 2024-02-19
Debug foca membership issues

spacekookie:


crates/corro-types/src/members.rs
```diff
@@ -46,41 +46,50 @@ pub struct Members {
...
             member.cluster_id = actor.cluster_id();
-
-            self.by_addr.insert(actor.addr(), actor.id());
-            self.recalculate_rings(actor.addr());
+            ret = MemberAddedResult::Updated;
         }
 
-        (newer, same)
+        self.recalculate_rings(actor.addr());
+        self.by_addr.insert(actor.addr(), actor.id());
```
jeromegn, line 90: Only need to do this if `NewMember`.

PR #162, opened 2024-03-01
Small subscription improvements

spacekookie:


crates/corro-types/src/pubsub.rs
```diff
@@ -299,6 +312,17 @@ impl MatcherHandle {
         info!(sub_id = %self.inner.id, "Canceled subscription");
     }
 
+    pub async fn cleanup_on_disk(self) {
+        info!(sub_id = %self.inner.id, "Attempting to cleanup...");
+        block_in_place(|| {
+            if let Err(e) = std::fs::remove_dir_all(&self.path) {
+                error!(sub_id = %self.inner.id, "could not delete subscription base path {} due to: {e}", self.path);
+            }
+        });
```
jeromegn, line 321: Should this just be using `tokio::fs` and not be in a `block_in_place`? I know `tokio::fs` operations block underneath, but that's handled properly internally.

PR #163, opened 2024-03-01
Count of member states should be more reliable?

jeromegn:
Somehow this could overflow a Vec capacity. The `SocketAddr` has size of 32, so it would have to be pretty enormous to fail, this is confusing.

If I understand correctly, a Vec can hold up to `isize::MAX` bytes. `isize::MAX / 32` is enormous...


jeromegn:
This has been merged as part of #164

PR #170, opened 2024-03-14
Derive Hash for RowId

pborzenkov:
So it can act as a key in a hash map.


jeromegn (approved):

spacekookie (approved):
Wow I don't think I've ever seen rustfmt break up a derive like _this_ before :joy:

PR #172, opened 2024-03-16
Fix deploy-on-fly

pnasrat:
Changes Dockerfile to use bookworm not latest for rust builder.

Updates docs to not checkout v0.1.0

Fixes ##171

Tested (destroyed the previous machine and volume and followed through doc launch steps after these changes)


spacekookie (approved):
Hey! Thank you for your contribution. This looks good to me.
I'll let CI run and then we can probably merge this

PR #176, opened 2024-03-26
Possible race in PG endpoint's commit handling

jeromegn:
When an explicit transaction (started with a `BEGIN` from the client) ends, it should be rolled back. This is not currently the case when an error occurs, for example.


jeromegn:
We should fix this at some point. IIRC it caused some issues when I tested it in development, but in general the way it works makes more sense than before. Must've introduced a bug.

crates/corro-pg/src/lib.rs
```diff
@@ -333,92 +334,87 @@ fn parse_query(sql: &str) -> Result<VecDeque<ParsedCmd>, ParseError> {
...
     }
 
-    fn end(&mut self) -> Option<OwnedSemaphorePermit> {
-        let permit = match self {
-            TxState::Started { write_permit, .. } => write_permit.take(),
-            TxState::Ended => None,
-        };
-        *self = TxState::Ended;
-        permit
+    fn set_failed(&mut self) {
```
gorbak25, line 398: Do we want to release the underlying write semaphore at this point so somebody else can use the connection?

gorbak25:
Superseded by #520

PR #177, opened 2024-03-26
Allow clearing empty versions via the CLI

spacekookie:


jeromegn (approved):
That'll do. We can add more info later on!

PR #181, opened 2024-04-03
corro-client: add CorrosionPooledClient

pborzenkov:
The pooled client is configured with multiple Corrosion servers and
falls back to alternative servers in case the current one can no longer
be connected to.

Otherwise, it tries to stick with a single server.


jeromegn (approved):
Looking good. I'd feel more comfortable with tests, but we can adjust as we go.

PR #189, opened 2024-04-12
Do not load _everything_ in memory for bookkeeping

jeromegn:
This adds some more data in sqlite to keep memory usage down and reduce time-to-start-syncing considerably.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1034,6 +1034,7 @@ dependencies = [
  "seahash",
  "serde",
  "serde_json",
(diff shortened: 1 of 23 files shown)


jeromegn (commented):
I'm a little concerned that Corrosion relies on perfect knowledge in bookkeeping to respond to other nodes. It's not different than before, but it is scary.

jeromegn:
Converting to draft because we should persist the gaps as soon as we detect a gap.

PR #195, opened 2024-04-23
Allow API listener to bind to multiple addresses

spacekookie:


crates/corro-types/src/config.rs
```diff
@@ -116,6 +116,8 @@ impl DbConfig {
 pub struct ApiConfig {
     #[serde(alias = "addr")]
     pub bind_addr: SocketAddr,
+    #[serde(alias = "extra_addrs")]
+    pub extra_bind_addrs: Vec<SocketAddr>,
```
jeromegn, line 120: `bind_addr` could be either a single address or multiple addresses. A `tokio::net::TcpListener::bind` call accepts multiple addresses. Might be easy with `serde_with`: https://docs.rs/serde_with/3.7.0/serde_with/struct.OneOrMany.html

jeromegn (changes requested):
It's a solution, but I think this could be a lot simpler

jeromegn (approved):

PR #208, opened 2024-05-14
 Properly calculate max version in stress_test

somtochiama:
This pull request calculates the max version for each node correctly at the end of stress tests and adds more logs to make it easier to figure out what caused the stress test to fail

### DIFF
--- a/crates/corro-agent/src/agent/tests.rs
+++ b/crates/corro-agent/src/agent/tests.rs
@@ -6,7 +6,7 @@ use std::{
 };
 
 use axum::Extension;
-use futures::{stream::FuturesUnordered, StreamExt, TryStreamExt};
+use futures::{future, stream::FuturesUnordered, StreamExt, TryStreamExt};
 use hyper::StatusCode;
 use rand::{
     distributions::Uniform, prelude::Distribution, rngs::StdRng, seq::IteratorRandom, SeedableRng,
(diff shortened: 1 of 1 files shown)


crates/corro-agent/src/agent/tests.rs
```diff
@@ -412,15 +412,20 @@ pub async fn configurable_stress_test(
...
     .try_buffer_unordered(10)
-    .try_collect::<BTreeMap<_, _>>()
-    .await?;
+    .try_collect::<Vec<(ActorId, usize)>>().await?
+    .into_iter().fold(BTreeMap::new(), |mut acc, item| {
+        {
+            *acc.entry(item.0).or_insert(0) += item.1
+        }
+        acc
+    });
```
sgrif, line 426:
```suggestion
    .try_fold(BTreeMap::new(), |mut acc, item| {
        {
            *acc.entry(item.0).or_insert(0) += item.1
        }
        Ok(acc)
    });
```

sgrif: I don't the amount of braces and derefs should be necessary here but I can't suggest more without seeing the compiler errors (happy to pair on this with you but it's not something particularly important)

sgrif (approved):
This looks fine, but a more detailed description of what you're changing and why this fixes it in the commit message and/or PR description would be helpful 😄

jeromegn (approved):

PR #214, opened 2024-05-23
Add function to handle subscriptions when applying buffered changes

jeromegn:
We discovered we weren't processing subscriptions for changes being applied all at once when all buffered versions were received. This caused some changes not to update subscriptions.

This PR adds a function to handle exactly that situation.

It's a bit painful and some buffering in memory is required unless we change the logic more extensively.


jeromegn:
@somtochiama I would appreciate a review on this. I think this works as intended and will prevent heartburn in the future.

somtochiama (approved):
LGTM!

PR #220, opened 2024-06-10
Send empties during sync based on previous timestamp

somtochiama:
This pull requests updates the sync process between nodes to also send previously cleared versions that the client node might have already received. Each time a node stores an empty version, it also stores a timestamp and can use this to know which versions have been cleared or updated since the last sync.


crates/corrosion/src/command/consul/sync.rs
```diff
@@ -705,7 +768,15 @@ mod tests {
...
+            "node-1",
+            &ta1_client,
+            update_services(HashMap::new(), &svc_hashes, false),
+            &mut svc_hashes,
+            Default::default(),
+            &mut check_hashes,
+        )
+        .await?;
 
         assert!(check_applied.is_zero());
```
jeromegn, line 782: Do I need to review this file or is it just formatting? It's fine to include this, but I was wondering 😅

somtochiama: I think this file is just formatting :sweat_smile:. I haven't figured out why format on save is slow on IDEA so I just run `cargo fmt` before I push

crates/corro-types/src/change.rs
```diff
@@ -232,12 +232,29 @@ pub fn insert_local_changes(
...
+            let ts = Timestamp::from(agent.clock().new_timestamp());
+            let inserted = store_empty_changeset(tx, actor_id, versions, ts)?;
+            if inserted > 0 {
+                cleared_ts = Some(ts)
+            }
         }
     }
 
+    if let Some(ts) = cleared_ts {
+        snap.update_cleared_ts(tx, ts)
```
jeromegn, line 250: just to be sure: this is updating the sync table, but for the current node. so I suppose it's useful if we do a snapshot so the restoring node is already up to date for this node.

crates/corro-agent/src/agent/util.rs
```diff
@@ -648,26 +658,47 @@ pub async fn process_fully_buffered_changes(
...
+                if inserted > 0 {
+                    last_cleared = Some(ts);
+                }
             }
         }
 
+        let mut agent_snap = agent_booked.snapshot();
+        if let Some(ts) = last_cleared {
+            agent_snap
+                .update_cleared_ts(&tx, ts)
```
jeromegn, line 678: I don't think we want to store the last cleared ts **except** during the sync process with the originating node.

crates/corro-agent/src/agent/util.rs
```diff
@@ -894,23 +942,100 @@ pub async fn process_multiple_changes(
...
+                    };
+
+                    let booked_write = booked.blocking_write(format!(
+                        "process_multiple_changes(booked writer, handling requested empties):{actor_id}",
+                    ));
+                    booked_write.snapshot()
+                }
+            };
+
+            snap.update_cleared_ts(&tx, max_ts)
```
jeromegn, line 990: same comment as before. we only want to store that if it's during sync with the originating node

crates/corro-agent/src/agent/util.rs
```diff
@@ -894,23 +942,100 @@ pub async fn process_multiple_changes(
...
+            })?;
+            snapshots.insert(*actor_id, snap);
+        }
+
+        let mut booked_writer = agent
+            .booked()
+            .blocking_write("process_multiple_changes(update_cleared_ts)");
+        let mut snap = booked_writer.snapshot();
+        if let Some(ts) = last_cleared {
+            snap.update_cleared_ts(&tx, ts)
```
jeromegn, line 1020: 🤔 isn't that redundant?

crates/corro-agent/src/api/peer.rs
```diff
@@ -702,6 +711,53 @@ fn handle_need(
...
+            let last_cleared_ts = last_cleared_ts.unwrap();
+            let mut filter = String::from("");
+            if let Some(ts) = ts {
+                if ts >= last_cleared_ts {
+                    warn!("received empties timestamp {ts:?} greater our last cleared ts: {last_cleared_ts:?}");
+                    return Ok(());
+                }
+
+                filter = format!("ts > {ts}");
+            }
```
jeromegn, line 727: I think `filter` could be avoided here. it doesn't matter much, but you could just trust the `ts` supplied by the requesting node and always filter like `ts > {ts}`. worse case: no results. I think the logic would be simpler. `ts` could default to `0`.

crates/corro-agent/src/api/peer.rs
```diff
@@ -702,6 +711,53 @@ fn handle_need(
...
+
+            let query = format!("SELECT start_version, end_version, ts from __corro_bookkeeping WHERE actor_id = ? AND end_version IS NOT NULL ORDER BY ts {filter}");
```
jeromegn, line 729: We might need an index on `__corro_bookkeeping.ts ASC` (maybe with `actor_id` too?). probably want to run a `EXPLAIN QUERY PLAN` on that to see what it uses.

jeromegn: You can also use `crsql_site_id()` instead of manually passing the current node's actor_id

crates/corro-types/src/agent.rs
```diff
@@ -350,6 +351,23 @@ fn refactor_corro_members(tx: &Transaction) -> rusqlite::Result<()> {
     "#,
     )
 }
+fn create_sync_state(tx: &Transaction) -> rusqlite::Result<()> {
+    tx.execute_batch(
+        r#"
+        CREATE TEMP TABLE _variables (name TEXT PRIMARY KEY, var TEXT);
+        INSERT INTO _variables VALUES ('current_ts', strftime('%Y-%m-%dT%H:%M:%S.000000000Z', CURRENT_TIMESTAMP));
+
+        UPDATE __corro_bookkeeping SET ts = (SELECT var FROM _variables WHERE name = 'current_ts') WHERE ts IS NULL;
```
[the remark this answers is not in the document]
jeromegn: also: should this only update the current node? the current node is **the source of truth for its own clears**.

jeromegn (commented):
Added a few comments, but the big possible problem I see here is: losing the information that we are missing some cleared versions.

If we keep updating the "last cleared ts" on every change we apply, then we don't know if we might've missed a cleared versions message in between.

I think this whole mechanism only works if we only store that information when we synchronize with another node and we store their cleared timestamp specifically. that way, we know for sure we've seen all versions the last time we synced with them.

Furthermore, the timestamps stored in bookkeeping help a bit because it gives us a way to reduce the synchronization bandwidth and processing load in the future. But these timestamps need to keep the original cleared timestamp, they can't be created from the current actor's clock.

In summary:
- if we keep updating the sync table with timestamps as we get them from any node, then we'll miss updates
- we should only update the last cleared ts when we sync and only for the node we're syncing with
- we should keep the original timestamp of the cleared version

crates/corro-types/src/agent.rs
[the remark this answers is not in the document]
somtochiama: > Interesting! Is that faster than just using strftime directly instead of a temporary table?

I don't think so. I am using a temp table because I want to put the same value in `__corro_sync_state` for this note. I guess its a big deal if they are slightly diffferent

crates/corro-agent/src/agent/util.rs
somtochiama: Here, we are updating the node's own last cleared ts. Some lines before, we check for overwritten version in the `__corro_versions_impacted` table and store empty changes. This table only contains cleared versions for that node and I have added a warning if for some reason it does not.

crates/corro-agent/src/agent/util.rs
somtochiama: It only sends a `SyncNeed::Empty` to a node's that its sync'ing with and each of those nodes sends their empties. But you are right, will add a check when it receives the empty changeset in `parallel_sync`

crates/corro-agent/src/agent/util.rs
somtochiama: the one above if for the received empties. this one is for the node's own empties (since we check for the node's overwrittten versions and we clear them in this function)

crates/corro-agent/src/api/peer.rs
```diff
@@ -609,11 +616,13 @@ fn handle_need(
                             last_seq,
                             ts,
                         )? {
+                            // todo: check if we have cleared the row locally
                             sender.blocking_send(SyncMessage::V1(SyncMessageV1::Changeset(
                                 ChangeV1 {
                                     actor_id,
                                     changeset: Changeset::Empty {
                                         versions: empty..=empty,
+                                        ts: Timestamp::from(agent.clock().new_timestamp()),
```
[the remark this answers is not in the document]
somtochiama: this is one of those cases where an empty version is discovered because there are no changes in the crsql_changes. It hasn't cleared it in its bookkeeping table, there's a check for non-null end versions before this section. (we could take the ts in the db anyways even though its not cleared)

crates/corro-agent/src/agent/util.rs
somtochiama: added the check!

crates/corro-types/src/change.rs
somtochiama: yes, it updating the sync for the current node.

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -401,6 +404,131 @@ fn processing_cost(change: &Changeset) -> usize {
...
+
+    let tx = conn
+        .immediate_transaction()
+        .map_err(|source| ChangeError::Rusqlite {
+            source,
+            actor_id: None,
+            version: None,
+        })?;
+
+    counter!("corro.sync.empties.count", "actor" => format!("{actor_id}"))
```
jeromegn, line 502: `actor_id.to_string()`

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -401,6 +404,131 @@ fn processing_cost(change: &Changeset) -> usize {
...
+    let mut max_ts: Timestamp = Default::default();
+    for (versions, ts) in changes {
+        debug!(self_actor_id = %agent.actor_id(), "processing emptyset changes, len: {}", versions.len());
+        for version in versions {
+            store_empty_changeset(&tx, actor_id, version.clone(), ts)?;
+        }
+        if ts > max_ts {
+            max_ts = ts;
+        }
+    }
```
jeromegn, line 513: should we use one transaction per emptyset? this would make the transactions much smaller and easier to digest for the database.

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -401,6 +404,131 @@ fn processing_cost(change: &Changeset) -> usize {
...
+            },
+            _ = &mut tripwire => {
+                break;
+            }
+        }
+
+        if process {
+            let mut retain_keys = Vec::new();
+            for (actor, changes) in &buf {
+                match process_emptyset(agent.clone(), bookie.clone(), *actor, changes.clone()).await
```
jeromegn, line 453: since you're awaiting without spawning, you can pass references instead of cloning.

crates/corro-agent/src/agent/handlers.rs
somtochiama: that makes sense.

somtochiama:
superseded by #223

PR #225, opened 2024-06-26
Insert missing hashes of consul services at start of consul sync

somtochiama:


crates/corrosion/src/command/consul/sync.rs
```diff
@@ -195,6 +196,123 @@ async fn setup(
...
+    let mut checks_insert = vec![];
+
+    for svc in services {
+        if service_hashes.get(&svc.id).is_none() {
+            let hash = hash_service(&svc);
+            statements.push(Statement::WithParams(
+                "INSERT INTO __corro_consul_services ( id, hash )
+    VALUES (?, ?)
+    ON CONFLICT (id) DO UPDATE SET
+        hash = excluded.hash;"
```
jeromegn, line 269: Probably easier to use `INSERT OR REPLACE INTO` instead of a query with an `ON CONFLICT`, given we always want to replace.

jeromegn (approved):

PR #229, opened 2024-07-02
pubsub: make sure to catch up sub in a transaction

pborzenkov:
So the last sent row and the last change ID values are consistent.
Otherwise it might skip all the changes happened between reading the rows
and querying the last change ID.


pborzenkov:
Not tested yet.

jeromegn (approved):

PR #230, opened 2024-07-02
Remove changes that failed to process from seen map

somtochiama:


crates/corro-agent/src/agent/handlers.rs
```diff
@@ -447,11 +450,20 @@ pub async fn handle_changes(
...
+            let agent = agent.clone();
+            let bookie = bookie.clone();
+            let seen_mutex = seen_mutex.clone();
+            join_set.spawn(async move {
+                if let Err(e) = util::process_multiple_changes(agent, bookie, changes.clone()).await
+                {
+                    error!("could not process multiple changes: {e}");
+                    for change in changes {
+                        let v = *change.0.versions().start();
+                        seen_mutex.lock().remove(&(change.0.actor_id, v));
```
jeromegn, line 463: Adding a mutex here is probably a bit harder than changing the return value of the future and handling it in the loop, in the `tokio::select!`. 

For example, you could return `BTreeMap<ActorId, RangeInclusiveSet<Version>>` (or something) from the future. It would represent the data to remove from the seen map. You'd only do this if the function errored, else you return an empty `BTreeMap` (does not allocate).

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -461,11 +473,8 @@ pub async fn handle_changes(
 
             // process these first, we don't care about the result,
             // but we need to drain it to free up concurrency
-            res = join_set.join_next(), if !join_set.is_empty() => {
+            _ = join_set.join_next(), if !join_set.is_empty() => {
```
jeromegn, line 476: Here, you'd handle the map, remove every element present in it, from `seen`.

jeromegn (changes requested):
Replacing the Mutex with simpler logic should make a smaller, more performant, code change.

jeromegn (approved):

PR #238, opened 2024-07-16
Truncate WAL if size above threshold

jeromegn:
Running `PRAGMA wal_checkpoint(TRUNCATE)` can be pretty expensive. Instead of running it on a timer, running it if its size is above a certain threshold makes more sense.

This should be made configurable..


somtochiama (approved):
LGTM

PR #245, opened 2024-07-26
Store and use last seen empty timestamp when requesting for empties

somtochiama:
When a node processes empties slowly, we can end up requesting already received empties that are still in the buffer. This pull request adds a field to the Members struct to store the last received empties timestamp and use that instead when requesting so we would not request an empty that we already have.

### DIFF
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -272,24 +272,7 @@ pub async fn handle_notifications(
         trace!("handle notification");
         match notification {
             Notification::MemberUp(actor) => {
-                // check for last empty when adding a new member.
-                let last_cleared_ts = {
-                    match agent.pool().read().await {
-                        Ok(conn) => match get_last_cleared_ts(&conn, actor.id()) {
-                            Ok(ts) => ts,
-                            Err(e) => {
-                                error!("could not get last_empty_ts: {e}");
-                                None
-                            }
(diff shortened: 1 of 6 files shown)


crates/corro-agent/src/agent/handlers.rs
```diff
@@ -273,7 +272,24 @@ pub async fn handle_notifications(
...
+                                error!("could not get last_empty_ts: {e}");
+                                None
+                            }
+                        },
+                        Err(e) => {
+                            error!("could not get read conn: {e}");
+                            None
+                        }
+                    }
+                };
```
jeromegn, line 290: It might be best to check the `member_added_res` and only query the DB if it is a newly added member (and then using `update_last_empty`).

The way SWIM works (or at least, `foca`), this can be called fairly often. I'll admit I don't remember if it's often enough to require moving this logic around, but I feel like changing it would be more consistent with how we usually manage the members list.

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -34,7 +34,6 @@ use corro_types::agent::ChangeError;
 use corro_types::base::Version;
 use corro_types::broadcast::Timestamp;
 use corro_types::change::store_empty_changeset;
-use corro_types::sync::get_last_cleared_ts;
```
jeromegn, line 37: Is that function still useful?

PR #246, opened 2024-07-29
corro-client: handle deserialization error correctly

pborzenkov:
Change ID still needs to be incremented even if the entry is not valid.


pborzenkov:
@jeromegn I've decided to deserialize as before and only fallback to non-typed `QueryEvent` in case deserialization fails, so we only pay the double deserialization price in these rare cases.

jeromegn (approved):

PR #248, opened 2024-07-30
cli/cluster members: send success reponse to close admin conn properly

runsisi:
otherwise admin command `corrosion cluster members` never finishes


jeromegn:
Good one! I noticed that too but forgot to fix it. Thank you.

PR #249, opened 2024-07-31
manually drop snapshot without running its destructor

runsisi:
this fixes the panic in debug build:

  thread 'tokio-runtime-worker' panicked at crates/corro-types/src/agent.rs:1306:9:
  max value was not applied

steps to reproduce:

  - schema

    CREATE TABLE todos ( id INTEGER NOT NULL PRIMARY KEY, title TEXT NOT NULL DEFAULT '' );

  - node A

    ./corrosion -c corro.toml exec 'INSERT INTO todos (id, title) VALUES (1, "test todo")'

  - node B

    ./corrosion -c corro.toml query 'SELECT * FROM todos' ./corrosion -c corro.toml exec 'DELETE FROM todos WHERE id == 1'

  now node A crashes with assert message `max value was not applied`

### DIFF
--- a/crates/corro-agent/src/agent/util.rs
+++ b/crates/corro-agent/src/agent/util.rs
@@ -976,6 +976,8 @@ pub async fn process_multiple_changes(
                     actor_id: None,
                     version: None,
                 })?;
+
+            std::mem::forget(snap);
         }
 
         tx.commit().map_err(|source| ChangeError::Rusqlite {


crates/corro-agent/src/agent/util.rs
```diff
@@ -962,20 +962,21 @@ pub async fn process_multiple_changes(
...
-
+            let mut snap = booked.snapshot();
             snap.update_cleared_ts(&tx, ts)
                 .map_err(|source| ChangeError::Rusqlite {
                     source,
                     actor_id: None,
                     version: None,
                 })?;
+
+            booked.commit_snapshot(snap);
```
jeromegn, line 979: Running this before we know the transaction succeeds can leave the in-memory bookkeeping in a desynchornized state, compared to the database.

jeromegn (changes requested):
The fix for the panic might be to `std::mem::forget(snap)`. it's a little weird all this all worked out, but we couldn't hold a write lock across the commit or else we'd keep the lock for far too long and some of our tests started failing.

crates/corro-agent/src/agent/util.rs
runsisi: got it. thanks!

jeromegn (approved):

PR #252, opened 2024-08-16
Drop items when processing queue is full

somtochiama:


jeromegn (approved):

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -838,6 +839,15 @@ pub async fn handle_changes(
             continue;
         }
 
+        // drop items when the queue is full.
+        if queue.len() > max_queue_len {
+            warn!(
+                "dropping changes from {} because changes queue is full",
+                change.actor_id
+            );
```
jeromegn, line 847: I've approved, but there are situations where this could spam the logs **a lot**

PR #254, opened 2024-08-23
Deduplicate needs requested during sync

somtochiama:


crates/corro-types/src/sync.rs
```diff
@@ -85,8 +85,72 @@ pub struct SyncStateV1 {
...
+                match n {
+                    SyncNeedV1::Full { versions } => {
+                        self.merge_full_version(*actor, versions);
+                    }
+                    SyncNeedV1::Partial { version, seqs } => {
+                        let mut delete = false;
+                        self.partial_need.entry(*actor).and_modify(|e| {
+                            if let Some(need_seqs) = e.get_mut(version) {
+                                let mut seqs_set =
+                                    RangeInclusiveSet::from_iter(need_seqs.clone().into_iter());
```
jeromegn, line 112: Why not use `need_seqs` directly here? You wouldn't have to clone these twice if you mutate the entry directly.

crates/corro-types/src/sync.rs
```diff
@@ -85,8 +85,72 @@ pub struct SyncStateV1 {
...
+                let range = *head + 1..=*version.start() - 1;
+                self.need.entry(actor_id).or_default().push(range);
+            }
+            *head = *version.end();
+        }
+
+        self.need.entry(actor_id).and_modify(|e| {
+            let mut set = RangeInclusiveSet::from_iter(e.clone().into_iter());
+            set.remove(version.clone());
+            *e = Vec::from_iter(set.into_iter());
```
jeromegn, line 150: Same as the other comment: you should be able to mutate the entry directly instead of cloning and re-creating range inclusives.

somtochiama:
Closed in favor of https://github.com/superfly/corrosion/pull/255

PR #257, opened 2024-08-29
keep gossip address the same as the actor address

runsisi:


jeromegn:
Is that causing issues?

runsisi:
we store actor address (`self.external_addr().unwrap_or_else(|| self.gossip_addr())`)[^1] in `__corro_members` table[^2], while in `spawn_swim_announcer` to skip local node announce[^3] gossip address (`gossip_server_endpoint.local_addr()`)[^4] is used as local address, i.e., `our_addr`. i think this is not correct.

maybe we should change `generate_bootstrap` [^5] signiture to ignore skip both gossip address and external address?

[^1]: https://github.com/superfly/corrosion/blob/e35acab3f3d6bb5499104ddf3b3dda7245a2437a/crates/corro-types/src/agent.rs#L141
[^2]: https://github.com/superfly/corrosion/blob/e35acab3f3d6bb5499104ddf3b3dda7245a2437a/crates/corro-agent/src/broadcast/mod.rs#L709
[^3]: https://github.com/superfly/corrosion/blob/e35acab3f3d6bb5499104ddf3b3dda7245a2437a/crates/corro-agent/src/agent/bootstrap.rs#L39
[^4]: https://github.com/superfly/corrosion/blob/e35acab3f3d6bb5499104ddf3b3dda7245a2437a/crates/corro-agent/src/agent/run_root.rs#L56
[^5]: https://github.com/superfly/corrosion/blob/e35acab3f3d6bb5499104ddf3b3dda7245a2437a/crates/corro-agent/src/agent/bootstrap.rs#L16

somtochiama:
> maybe we should change generate_bootstrap [5](https://github.com/superfly/corrosion/pull/257#user-content-fn-5-cd1d06b2aaf3c58039e64195db284924) signiture to ignore skip both gossip address and external address?

@runsisi I think this is a better solution.

PR #262, opened 2024-10-22
Add endpoint to receive updates for tables 

somtochiama:
This pull request adds a HTTP endpoint `/updates/:table` that sends future updates about the primary keys of rows have been updated or deleted. 

```
$ curl -X POST http://localhost:8081/v1/updates/todos
{"notify":["upsert",["some-id-82"]]}
{"notify":["upsert",["1"]]}
{"notify":["delete",["1"]
```


crates/corro-api-types/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub enum TypedQueryEvent<T> {
     },
     Change(ChangeType, RowId, T, ChangeId),
     Error(CompactString),
+    Notify(ChangeType, T)
```
jeromegn, line 38: I don't think we should reuse this enum. The updates manager is always going to be sending `Notify` and the subs manager will never be sending `Notify`. Re-using the same enum is more complicated for clients because now they have to handle an enum variant that'll never matcher anyway.

crates/corro-types/src/pubsub.rs
```diff
@@ -166,153 +187,14 @@ impl SubsManager {
...
 #[derive(Debug)]
-struct MatchableChange<'a> {
-    table: &'a TableName,
-    pk: &'a [u8],
-    column: &'a ColumnName,
+pub struct MatchableChange<'a> {
+    pub table: &'a TableName,
+    pub pk: &'a [u8],
+    pub column: &'a ColumnName,
+    pub cl: &'a i64,
```
jeromegn, line 197: I think using a `i64` directly instead of a reference is best here. I believe they have the same size and it's just easier to deal with integers directly than references to them imo.

crates/corro-types/src/pubsub.rs
```diff
@@ -384,19 +267,72 @@ struct InnerMatcherHandle {
     cached_statements: HashMap<String, MatcherStmt>,
 }
 
-type MatchCandidates = IndexMap<TableName, IndexSet<Vec<u8>>>;
+pub type MatchCandidates = IndexMap<TableName, IndexSet<(Vec<u8>, i64)>>;
```
jeromegn, line 270: This is now pretty expensive to lookup against. Previously we could just pass a slice, but now we have to convert to a `Vec<u8>` so we can check `&(vec, cl)`.

Could this be an associated type on the `Handle` trait?

The subs could use `IndexMap<TableName, IndexSet<Vec<u8>>>` and the updates could use `IndexMap<TableName, IndexMap<Vec<u8>, i64>>` and then you wouldn't have to allocate every time you want to check if it's contained.

An associated type looks like:

```rust
trait Handle {
  type CandidatesMatcher;
  // ...
}
```

Then every implementation needs to define that. I think it might work!

crates/corro-types/src/updates.rs
```diff
@@ -0,0 +1,501 @@
...
+        }
+    }
+
+    return Ok(());
+}
+
+async fn cmd_loop(
+    id: Uuid,
+    cancel: CancellationToken,
+    evt_rx: mpsc::Sender<QueryEvent>,
```
jeromegn, line 279: Should be `evt_tx`.

jeromegn (changes requested):

crates/corro-api-types/src/lib.rs
somtochiama: I was thinking we might want to send `Columns` at the start of the updates but that could be put in the new enum too.

jeromegn: Yes, but we can also assume clients know the primary keys of a table, I think. On subscriptions, there can be a lot of columns and they can have generated names, but not for these.

crates/corro-types/src/updates.rs
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    // metrics...
+    for (id, (candidates, handle)) in candidates {
+        let mut match_count = 0;
+
+        for (table, pks) in candidates.iter() {
+            let count = pks.len();
+            match_count += count;
+            counter!(format!("corro.{trait_type}.changes.matched.count"), "sql_hash" => handle.hash().to_string(), "table" => table.to_string()).increment(count as u64);
```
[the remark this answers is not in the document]
somtochiama: hmm. this would probably require grabbing a lock before getting the `Counter` which is also not great.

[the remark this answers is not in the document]
jeromegn: You can pre-allocate them since you already know the sql hash and table names at creation time.

crates/corro-agent/src/api/public/update.rs
```diff
@@ -0,0 +1,284 @@
...
+                }
+
+                // reset the deadline if there are receivers!
+                deadline = None;
+                continue;
+            },
+            _ = subs_check.tick() => {
+                if tx.receiver_count() == 0 {
+                    if deadline.is_none() {
+                        deadline = Some(Box::pin(tokio::time::sleep(MAX_UNSUB_TIME)));
```
jeromegn, line 145: I don't think we need the whole deadline check because it's stateless. As soon as update listeners are gone, we should just remove it.

It should simplify the code significantly here.

crates/corro-types/src/updates.rs
```diff
@@ -0,0 +1,560 @@
...
+                };
+                handle.filter_matchable_change(candidates, change);
+            }
+        }
+    }
+
+    // metrics...
+    for (id, (candidates, handle)) in candidates {
+        let mut match_count = 0;
+        for (table, pks) in candidates.clone() {
```
jeromegn, line 524: Would `.iter()` be sufficient instead of `.clone()`? It doesn't look like you need a owned value in that loop.

crates/corro-types/src/updates.rs
```diff
@@ -0,0 +1,560 @@
...
+                .get_counter(&table)
+                .matched_count
+                .increment(pks.len() as u64);
+        }
+
+        trace!(sub_id = %id, %db_version, "found {match_count} candidates");
+
+        if let Err(e) = handle
+            .changes_tx()
+            .try_send((candidates.clone(), db_version))
```
jeromegn, line 537: if you get rid of the other clone, you can remove this one as well.

crates/corro-types/src/updates.rs
```diff
@@ -0,0 +1,560 @@
...
+        let mut candidates = MatchCandidates::new();
+        let mut match_count = 0;
+        for change in changes.iter().map(MatchableChange::from) {
+            if handle.filter_matchable_change(&mut candidates, change) {
+                match_count += 1;
+            }
+        }
+
+        // metrics...
+        for (table, pks) in candidates.clone() {
```
jeromegn, line 433: I believe `.iter()` works here as well.

jeromegn (commented):
Outlined a few changes that can help reduce complexity and make it more efficient. It's looking pretty good though!

crates/corro-types/src/updates.rs
somtochiama: ah. this was from when I had the associated type. will take it out.

jeromegn (approved):

PR #264, opened 2024-11-01
Prioritize recent versions during sync

somtochiama:


crates/corro-agent/src/agent/handlers.rs
```diff
@@ -863,15 +863,22 @@ pub async fn handle_changes(
             continue;
         }
 
-        // drop items when the queue is full.
+        // drop old items when the queue is full.
         if queue.len() > max_queue_len {
+            let change = queue.pop_front();
```
jeromegn, line 868: Should this be `pop_back`? We're popping from the front in the rest of the logic.

somtochiama: Ah yes. We are processing received changes in order since biggest versions sent first.

jeromegn (approved):

PR #269, opened 2024-11-28
Handle failed changes properly

somtochiama:
This pull request adds the following changes:
- The use of savepoints for commit single versions are handled within `process_multiple_changes` so if one version fails we can continue with the others. 
- It reverts the behaviour of removing failed versions from the seen map as this can cause duplicate changes from broadcasts to be re-sent. Now the changes will eventually be retried when the cache is cleared.
- We now maintain the same size for the both the cache and the queue, so that we never have duplicate items in the queue

### DIFF
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -750,8 +750,8 @@ pub async fn handle_changes(
         agent.config().perf.apply_queue_timeout as u64,
     ));
 
-    const MAX_SEEN_CACHE_LEN: usize = 10000;
-    const KEEP_SEEN_CACHE_SIZE: usize = 1000;
+    let max_seen_cache_len: usize = max_queue_len;
+    let keep_seen_cache_size: usize = cmp::max(10, max_seen_cache_len / 10);
     let mut seen: IndexMap<_, RangeInclusiveSet<CrsqlSeq>> = IndexMap::new();
 
     let mut drop_log_count: u64 = 0;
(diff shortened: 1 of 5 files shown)


crates/corro-agent/src/agent/handlers.rs
```diff
@@ -747,8 +750,8 @@ pub async fn handle_changes(
         agent.config().perf.apply_queue_timeout as u64,
     ));
 
-    const MAX_SEEN_CACHE_LEN: usize = 10000;
-    const KEEP_SEEN_CACHE_SIZE: usize = 1000;
+    let max_seen_cache_len: usize = max_changes_chunk;
```
jeromegn, line 753: Shouldn't that be `max_queue_len`?

crates/corro-agent/src/agent/tests.rs
```diff
@@ -1157,7 +1298,7 @@ async fn get_rows(
                 |row| row.get(0),
             )?;
             let mut last = 4;
-            // count will be zero for cleared versions
+            // count will be zero for cleared vertest_processions
```
jeromegn: Find and replace? :D

crates/corro-agent/src/agent/tests.rs
```diff
@@ -903,6 +908,106 @@ async fn test_clear_empty_versions() -> eyre::Result<()> {
...
+            .query_row([i], |row| row.get::<_, String>(0))?;
+    }
+
+    let conn = ta1.agent.pool().read().await?;
+    let res = conn
+        .prepare_cached("SELECT text from tests where id = 6")?
+        .query_row([], |row| row.get::<_, String>(0));
+    assert!(res.is_err());
+    assert_eq!(res, Err(rusqlite::Error::QueryReturnedNoRows));
+
```
jeromegn, line 1003: Can you also check that the `db_version` inserted and the ones Corrosion associated with its internal version are correct? Possibly by trying to apply the bad change first.

I vaguely recall we do some db version arithmetics and we might need to subtract when we're rolling back a savepoint.

jeromegn (commented):
Left a few comments.

crates/corro-agent/src/agent/tests.rs
somtochiama: 😅 Sometimes my thinkpad does some weird coy-paste thing when I use the touchpad. Need to disable it..

jeromegn (approved):

PR #279, opened 2024-12-20
Send cleared versions to follower

somtochiama:


somtochiama:
Old following depended on the total ordering provided by the db_version which is now gone, Closing this

PR #283, opened 2025-01-20
implement tls on corro-pg

lucas-jacques:


crates/corro-pg/src/lib.rs
```diff
@@ -472,26 +553,58 @@ pub async fn start(
...
-                let mut framed = Framed::new(
-                    conn,
-                    PgWireMessageServerCodec::new(ClientInfoHolder::new(remote_addr, false)),
-                );
+                let mut secured_conn = false;
+
+                let mut framed = {
+                    if ssl_supported && ssl {
+                        // safe to unwrap here because we checked for ssl_supported above
+                        let tls_acceptor = tls_acceptor.unwrap();
```
jeromegn, line 580: To prevent unwrapping, you could `match (ssl, tls_acceptor.take()) {` and the branch `(true, Some(tls_acceptor))` can handle the handshake.

Unwrapping (even when it's safe because we checked a precondition), is usually a sign we're working against the type system.

jeromegn (commented):
Looks pretty good! You could try a small refactor to prevent unwrapping the tls acceptor.

crates/corro-pg/src/lib.rs
lucas-jacques: Made an update [cf24dc2](https://github.com/superfly/corrosion/pull/283/commits/cf24dc242abf60b297652a80168401d5779490b4)

crates/corro-pg/src/lib.rs
```diff
@@ -464,6 +458,85 @@ pub enum PgStartError {
...
+    PgTlsErrro(#[from] eyre::Error),
+}
+
+async fn setup_tls(pg: PgConfig) -> eyre::Result<(Option<TlsAcceptor>, bool)> {
+    if pg.tls.is_none() {
+        return Ok((None, false));
+    }
+    let mut ssl_required = false;
+
+    let tls = pg.tls.unwrap();
```
jeromegn, line 471:
```rust
let tls = match pg.tls {
  Some(tls) => tls,
  None => {
    return Ok((None, false));
  }
};

let ssl_required = tls.verify_client;
```

jeromegn: Would've been nicer with a more recent rustc, but we don't have that right now (we do in a branch, but we can't just upgrade in this one).

jeromegn (commented):
Looks good, I made a small comment to prevent an `unwrap`.

crates/corro-pg/src/lib.rs
lucas-jacques: oops, i missed this one

crates/corro-pg/src/lib.rs
```diff
@@ -464,6 +458,90 @@ pub enum PgStartError {
     Io(#[from] std::io::Error),
     #[error(transparent)]
     Rusqlite(#[from] rusqlite::Error),
+    #[error(transparent)]
+    PgTlsErrro(#[from] eyre::Error),
```
somtochiama, line 462: Small spelling fix
```suggestion
    PgTlsError(#[from] eyre::Error),
```

crates/corro-pg/src/lib.rs
```diff
@@ -464,6 +458,90 @@ pub enum PgStartError {
...
+}
+
+async fn setup_tls(pg: PgConfig) -> eyre::Result<(Option<TlsAcceptor>, bool)> {
+    if pg.tls.is_none() {
+        return Ok((None, false));
+    }
```
somtochiama, line 468: I think this is no longer needed with the match statement below.
```suggestion
```

somtochiama (commented):
Thanks for making this PR @lucas-jacques

This looks good! Just a couple of nitpics

jeromegn:
Hmm, sorry, I merged another branch that was more important and now yours has conflicts. Would you mind fixing them?

lucas-jacques:
@jeromegn No problem, it should be ok now

PR #286, opened 2025-02-05
Timeout sqlite operations

somtochiama:


jeromegn (approved):
looks pretty good! I think the default timeout is too long, but we can adjust.

crates/corro-types/src/agent.rs
```diff
@@ -804,6 +820,12 @@ impl DerefMut for WriteConn {
     }
 }
 
+impl Drop for WriteConn {
+    fn drop(&mut self) {
+        self.cancel.cancel();
+    }
+}
```
jeromegn, line 827: DropGuard was probably the better choice here.

PR #301, opened 2025-04-30
Increase WAL checkpoint timeout

PeterCxy:
busy_handler of checkpoints will wait until all readers are on the same latest snapshot, NOT when all readers complete. With huge subs we often have readers that take very long to complete, and in that case we might keep missing checkpoints until the WAL blows up in size, slowing down readers even further.

Because we don't actually need to wait for all readers to fully complete, and it doesn't block new readers (they'll be on the new snapshot), try increasing the timeout by quite a lot. This will still cause writes (changes) to temporarily queue up, but at least this will _not_ be blowing everything up if a lot of slow readers (subs) constantly block checkpointing.


somtochiama:
Closed in favor of https://github.com/superfly/corrosion/pull/306/commits/6234b5534dc1fc52e08d8d50d92e7bdce44a98cc

PR #312, opened 2025-06-16
Add destructive flag to allow destructive changes

somtochiama:


somtochiama:
Enabling destructive changes requires more thought since it could lead to updates on existing rows in every node

Issue #310, opened 2025-05-31
Inconsistencies between code, comment and assertion message


YichiZhang0613:
In crates/backoff/src/lib.rs, the comment indicates function will panic when jitter < 0.0 or jitter > 1.0. But the code will panic when jitter <= 0.0 or jitter >= 1.0. While assertion message " jitter must be between 0 and 1." indicates that jitter must be >= 0.0 and <= 1.0.
```rust
/// ## Panics
    /// This method panics if a number smaller than `0` or larger than `1` is
    /// provided.
    #[inline]
    pub fn jitter(mut self, jitter: f32) -> Self {
        assert!(
            jitter > 0f32 && jitter < 1f32,
            "<exponential-backoff>: jitter must be between 0 and 1."
        );
        self.jitter = jitter;
        self
    }
```

fatkobra:
You have two options depending on desired behavior:

Keep current code (exclusive bounds), update doc and message
`/// ## Panics
/// This method panics if a number less than or equal to `0`, or greater than or equal to `1` is provided.
#[inline]
pub fn jitter(mut self, jitter: f32) -> Self {
    assert!(
        jitter > 0f32 && jitter < 1f32,
        "<exponential-backoff>: jitter must be strictly between 0 and 1 (exclusive)."
    );
    self.jitter = jitter;
    self
}
`

or Change code to allow 0.0 and 1.0 (inclusive), if that's the intended behavior
`/// ## Panics
/// This method panics if a number smaller than `0` or larger than `1` is provided.
#[inline]
pub fn jitter(mut self, jitter: f32) -> Self {
    assert!(
        jitter >= 0f32 && jitter <= 1f32,
        "<exponential-backoff>: jitter must be between 0 and 1 (inclusive)."
    );
    self.jitter = jitter;
    self
}
`

PR #319, opened 2025-06-22
Fix inconsistent comment in backoff crate

YichiZhang0613:
Fix problems mentioned in #310 .

### DIFF
--- a/crates/backoff/src/lib.rs
+++ b/crates/backoff/src/lib.rs
@@ -36,13 +36,12 @@ impl Backoff {
     /// Set the amount of jitter per backoff.
     ///
     /// ## Panics
-    /// This method panics if a number smaller than `0` or larger than `1` is
-    /// provided.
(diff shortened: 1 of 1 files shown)


crates/backoff/src/lib.rs
```diff
@@ -41,7 +41,7 @@ impl Backoff {
     #[inline]
     pub fn jitter(mut self, jitter: f32) -> Self {
         assert!(
-            jitter > 0f32 && jitter < 1f32,
+            jitter >= 0f32 && jitter <= 1f32,
```
somtochiama, line 44: I think it is the comment that is incorrect. Allowing a jitter of `0` would cause a panic when trying to because the range used in `thread_rng().gen_range(0..(jitter_factor * 2));` would be empty. I can't see a reason why 1 shouldn't be allowed but I am inclined to just leave the code for the two and just fix the comment

PR #335, opened 2025-08-07
Fix clippy lints

Jake-Shadle:
Follow on to #334 so that change wasn't larger, I can rebase if it is merged, or split it out if not.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,8 +114,4 @@ codegen-units = 16
 opt-level = 2
 
 [workspace.lints.clippy]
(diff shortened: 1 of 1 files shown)


somtochiama:
@Jake-Shadle I think most of these fixes got into #334 . Should we close this?

Jake-Shadle:
Oh no, I purposefully disabled some lints to make #334 smaller, so this removes them and fixes them.

somtochiama (approved):
Thanks @Jake-Shadle !

PR #337, opened 2025-08-12
Add support for multiple PostgreSQL endpoints and readonly mode

PeterCxy:
As the title says, this PR makes it possible to spawn multiple PostgreSQL server endpoints, some of which may be readonly. Readonly mode is implemented by setting the readonly flag on the underlying SQLite connection, so that there's absolutely no possibility of accidentally running a write transaction using a readonly server.


somtochiama (approved):
Thanks @PeterCxy !
LGTM!

PR #345, opened 2025-08-29
Update HTTP stack

Jake-Shadle:
Following from #342, this updates the bulk of the HTTP stack and all of the crates that were related them, so everything TLS related, etc.

* `axum` - 0.6 => 0.8 - This introduced slight changes to the startup of the server loop as axum removed helpers from axum itself.
* `hyper` - 0.14 => 1.7 - Fairy large change, several types/helpers/utilities that were previously in hyper itself have been moved out to `hyper-util`, `http-body-util` etc. Notably, though the stream Channel body was moved to `http-body-util`, the way it was used in this project couldn't be directly ported, so I've placed a minimal copy in corro-agent with the functionality used by this project, the ability to know when the receiver body has been closed during the periodic flushes to know when to drop the streaming sender.
* `hyper::Client` => `reqwest` - The hyper client code changed quite a bit, and considering this project's usage I just opted to change it to use reqwest instead, so still on the same stack, just slightly easier to work with.
* `rustls` - 0.21 => 0.23 - There was a lot of movement in the rustls crates. Biggest note is that rustls-pemfile, which this project used quite a bit, has been archived and won't get more updates, I've opted to use the re-export of rustls-pki-types from rustl::pki_types to convert all of the code that loads certs/keys. There was quite a lot of this code that was almost the same in several placed but I only converted in place instead of folding some of the functions into a single place. Also of note, 0.22 introduced the aws-lc-rs crate as an alternative to ring, I've configured features for all the crates that depend on rustls so that ring is still the default crypto provider. This is extremely finicky as some crates now default to using aws-lc-rs instead of ring (eg. tokio-rustls) which will cause code to break at runtime as having both aws-lc-rs and ring turned on will mean that rustls will not have a default crypto provider and basically all TLS related code will break.
* `quinn` - 0.10 => 0.11 - Nothing really big changed here, though SendStream::finish is no longer a future, and now must (should) be followed by a SendStream::stopped which is a future.
* `tracing-opentelemetry` - Trait shenanigans required this to updated otherwise HTTP could not be traced


crates/corro-agent/src/agent/handlers.rs
```diff
@@ -62,26 +62,39 @@ pub fn spawn_gossipserver_handler(
...
+                        break;
+                    }
+                    incoming = gossip_server_endpoint.accept() => {
+                        if let Some(inc) = incoming {
+                            inc.refuse();
+                        }
+                    }
+                }
+            }
+
```
Jake-Shadle, line 97: quinn has split up the connect stage more, giving more flexibility, so the `reject_new_connections` method was removed, as now the `Incoming` type allows connections to be `refuse`ed.

crates/corro-agent/src/agent/tests.rs
```diff
@@ -74,18 +72,15 @@ async fn insert_rows_and_gossip() -> eyre::Result<()> {
...
-                .method(hyper::Method::POST)
-                .uri(format!("http://{}/v1/transactions", ta1.agent.api_addr()))
-                .header(hyper::header::CONTENT_TYPE, "application/json")
-                .body(serde_json::to_vec(&req_body)?.into())?,
-        ),
+        client
+            .post(format!("http://{}/v1/transactions", ta1.agent.api_addr()))
+            .header(hyper::header::CONTENT_TYPE, "application/json")
+            .body(serde_json::to_vec(&req_body)?)
+            .send(),
```
Jake-Shadle, line 79: The difference between hyper::Client and reqwest are slight, but tend to result in smaller code.

crates/corro-agent/src/agent/tests.rs
```diff
@@ -623,18 +608,15 @@ async fn large_tx_sync() -> eyre::Result<()> {
...
-                hyper::Request::builder()
-                    .method(hyper::Method::POST)
-                    .uri(format!("http://{}/v1/transactions", ta1.agent.api_addr()))
-                    .header(hyper::header::CONTENT_TYPE, "application/json")
-                    .body(serde_json::to_vec(&req_body)?.into())?,
-            ),
+            client
+                .post(format!("http://{}/v1/transactions", ta1.agent.api_addr()))
+                .header(hyper::header::CONTENT_TYPE, "application/json")
+                .body(serde_json::to_vec(&req_body)?)
```
Jake-Shadle, line 614: I didn't do this to stay as close to the old code as possible, but this code could be simplified to use `.json(<Serialize>)` which automatically sets the content_type

crates/corro-agent/src/agent/util.rs
```diff
@@ -222,7 +227,7 @@ pub async fn setup_http_api_handler(
             ),
         )
         .route(
-            "/v1/updates/:table",
+            "/v1/updates/{table}",
```
Jake-Shadle, line 230: This is how path captures are done in axum now.

crates/corro-agent/src/agent/util.rs
```diff
@@ -288,38 +293,79 @@ pub async fn setup_http_api_handler(
...
-                    api.clone()
-                        .into_make_service_with_connect_info::<SocketAddr>(),
-                )
-                .with_graceful_shutdown(
-                    tripwire
-                        .clone()
-                        .inspect(move |_| info!("corrosion api http tripped {api_addr}")),
-                )
-                .inspect(|_| info!("corrosion api is done")),
-        );
```
Jake-Shadle, line 312: - hyper removed AddrIncoming
- axum removed Server::builder

crates/corro-agent/src/api/peer/mod.rs
```diff
@@ -340,25 +314,57 @@ pub async fn gossip_client_endpoint(config: &GossipConfig) -> eyre::Result<quinn
...
 impl SkipServerVerification {
-    fn new() -> Arc<Self> {
-        Arc::new(Self)
+    fn new(provider: Arc<rustls::crypto::CryptoProvider>) -> Arc<Self> {
+        Arc::new(Self(provider))
     }
 }
 
-impl rustls::client::ServerCertVerifier for SkipServerVerification {
+impl rustls::client::danger::ServerCertVerifier for SkipServerVerification {
```
Jake-Shadle, line 326: This trait is more complicated now, the new methods are just forwarded to the actual provider.

crates/corro-agent/src/api/peer/mod.rs
```diff
@@ -1304,9 +1310,12 @@ pub async fn parallel_sync(
                 }
 
                 if needs.is_empty() {
-                    if let Err(e) = tx.finish().instrument(info_span!("quic_finish")).await {
+                    if let Err(e) = tx.finish() {
                         warn!("could not finish stream while sending sync requests: {e}");
+                    } else if let Err(e) = tx.stopped().instrument(info_span!("quic_stopped")).await {
+                        warn!("could not wait for stream stopped while sending sync requests: {e}");
                     }
```
Jake-Shadle, line 1317: As stated in the description, quinn split finish into finish and stopped, the latter of which is now the future that needs to be waited on for any buffered sends to be flushed.

crates/corro-agent/src/api/public/pubsub.rs
```diff
@@ -842,8 +844,8 @@ async fn forward_bytes_to_body_sender(
                         return;
                     }
                 } else {
-                    if let Err(e) = poll_fn(|cx| tx.poll_ready(cx)).await {
-                        warn!(%sub_id, error = %e, "body sender was closed or errored, stopping event broadcast sends");
+                    if tx.is_closed() {
+                        warn!(%sub_id, "body sender was closed, stopping event broadcast sends");
```
Jake-Shadle, line 847: The Sender in http_body_util no longer has poll_ready or any other way to determine if the receiving body is closed.

crates/corro-agent/src/streaming_body.rs
```diff
@@ -0,0 +1,76 @@
+use http_body::Frame;
```
Jake-Shadle, line 1: Minimal copy of http_body_util's Channel body.

crates/corro-agent/src/streaming_body.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.tx_frame
+            .send(Frame::data(buf))
+            .await
+            .map_err(|_| SendError)
+    }
+
+    #[inline]
+    pub fn is_closed(&self) -> bool {
+        self.tx_frame.is_closed()
+    }
```
Jake-Shadle, line 67: The functionality I needed not in that crate.

crates/corro-pg/src/codec.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.protocol_version = version;
+    }
+
+    #[inline]
+    fn client_certificates<'a>(&self) -> Option<&[rustls::pki_types::CertificateDer<'a>]> {
+        None
+    }
+}
+
+pub struct PgWireMessageServerCodec {
```
Jake-Shadle, line 103: This type was (accidentally?) made private in a code reorganization, so I've copied it here to reduce the churn in the lib file.

crates/corro-pg/src/lib.rs
```diff
@@ -3378,598 +3352,3 @@ fn field_types(
 
     Ok(fields)
 }
-
-#[cfg(test)]
```
Jake-Shadle, line 3382: I moved all the tests to an integration test file, this single file is massive, and would mean iterating on test code would rebuild the library as well, even if the change only occurred in test code.

crates/corro-pg/src/ssl.rs
```diff
@@ -0,0 +1,96 @@
+use futures::{SinkExt, StreamExt};
```
Jake-Shadle, line 1: This is a copy of the pgwire SSL code which is also private.

crates/corro-pg/src/codec.rs
```diff
@@ -0,0 +1,171 @@
+//! Reimpl of PgWireMessageServerCodec from <https://github.com/sunng87/pgwire/blob/master/src/tokio/server.rs>,
```
somtochiama: can we pin the to a particular commit so the link works even if main changes

PR #350, opened 2025-09-08
corro-pg: ensure connections are closed on unexpected error

rianmcguirefly:
Ensure all the tasks we spawn when handling a postgresql connection exit before the [main connection handler task](https://github.com/superfly/corrosion/blob/179a07932da99233973e35e5cacf5e8dac505b23/crates/corro-pg/src/lib.rs#L566), so the TCP connection isn't ever left open.

I don't really know what I'm doing with Rust/Tokio, so there might be better ways to approach this.

### DIFF
--- a/crates/corro-pg/src/lib.rs
+++ b/crates/corro-pg/src/lib.rs
@@ -671,50 +671,64 @@ pub async fn start(
 
                 let cancel = CancellationToken::new();
 
-                tokio::spawn({
+                let frontend_future = tokio::spawn({
                     let back_tx = back_tx.clone();
                     let cancel = cancel.clone();
                     async move {
                         // cancel stuff if this loop breaks
                         let _drop_guard = cancel.drop_guard();
 
-                        while let Some(decode_res) = stream.next().await {
-                            let msg = match decode_res {
-                                Ok(msg) => msg,
-                                Err(PgWireError::IoError(io_error)) => {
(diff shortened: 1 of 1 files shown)


crates/corro-pg/src/lib.rs
```diff
@@ -671,50 +671,64 @@ pub async fn start(
...
+                                            );
+                                            break;
+                                        }
+                                    };
+
+                                    if front_tx.send(msg).await.is_err() {
+                                        break;
+                                    }
+                                }
+                                _ = front_tx.closed() => {
```
rianmcguirefly, line 720: If we're not going to process any more incoming messages, we can stop waiting to read them from `stream`.

The main `spawn_blocking` task drops `front_rx` when it completes, which causes this to become closed.

crates/corro-pg/src/lib.rs
```diff
@@ -1899,6 +1913,12 @@ pub async fn start(
                     }
                 }
 
+                // Close the backend channel once all messages have been sent
+                drop(back_tx);
+
+                let _ = backend_future.await;
+                let _ = frontend_future.await;
```
rianmcguirefly, line 1920: Ensure these tasks complete, otherwise they can hold `conn` open.

rianmcguirefly: Should we be more defensive here and abort the tasks after a timeout?

PR #351, opened 2025-09-08
Abort PG TCP connection loops when message-handling exits

PeterCxy:
Ideally we should have a wrapper that aborts these automatically on drop, but for now this should fix the bug at hand.

### DIFF
--- a/crates/corro-pg/src/lib.rs
+++ b/crates/corro-pg/src/lib.rs
@@ -1857,11 +1857,6 @@ pub async fn start(
                     }
                 }).await;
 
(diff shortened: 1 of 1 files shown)


PeterCxy:
This also aborts the connection if client-side decides to send a Terminate message, though according to PG docs this should be the expected behavior:

> The normal, graceful termination procedure is that the frontend sends a Terminate message and immediately closes the connection. On receipt of the message, the backend immediately closes the connection and terminates.

There's no mention of expecting to drain the TCP socket by performing a graceful shutdown there.

somtochiama (approved):
LGTM!

PR #358, opened 2025-09-16
Parse integers as uint first in `pack_columns` to prevent sign extension

somtochiama:
The `pack_column` function keeps all bytes for negative integers but truncates the zeros for positve integers. The [Bytes get_int](https://github.com/tokio-rs/bytes/pull/732) function previously used has been updated to do a sign extension for when the number of bits are less than 64, and this would cause some positive integers (with 1 as the msb) from the db to be changed to negative integers when unpacked. This pull request fixes it by doing a `uint` conversion which would add back the zeros.


FreeLife-Web (commented):
intresting🔥👾

PR #371, opened 2025-10-01
Add counters for pg + http writes

somtochiama:


somtochiama:
@gorbak25 fair point. We do have a counter for the number of write permits available (so it will show up) but we might not know it is pg causing this. Just pushed a commit.

PR #373, opened 2025-10-06
Track active TCP streams in the user facing API

gorbak25:
Introduces the `corro.api.active.streams` gauge measuring the amount of currently active data streams. 
In the HTTP api it's implemented by wrapping `hyper::Body` in an wrapper struct, in the postgres listener I wrapped `TcpStream`.

This gauge tells us how many active subscriptions and postgres connections we are currently handling.


PeterCxy:
You might want to "touch" the gauge periodically to prevent it from getting GC'd -- if that happens, any reference to the old gauge object will not do anything (so the connection drops won't be recorded) and anything new will restart from 0.

gorbak25:
@PeterCxy @somtochiama The "Touch the gauge periodically" pattern was repeated in the codebase. 

I've made an `persistent_gauge!` wrapper for the `gauge!` macro. Gauges registered via `persistent_gauge!` get always updated every 30s using `increment(0.0)`. I've changed `gauge!` to `persistent_gauge!` where appropriate.

somtochiama (approved):
LGTM! Thanks @gorbak25

PR #374, opened 2025-10-07
Add membership id for cluster membership

somtochiama:
⚠️ This is a breaking change as bincode (the library for encoding foca data) doesn't support default fields and the `Actor` struct has been updated with a new field `member_id`.

### DIFF
--- a/crates/corro-agent/src/agent/tests.rs
+++ b/crates/corro-agent/src/agent/tests.rs
@@ -1440,19 +1440,19 @@ async fn test_diff_member_id() -> eyre::Result<()> {
 
     let members = ta1.agent.members().read().states.clone();
     assert_eq!(members.len(), 1);
-    assert!(members.get(&ta3.agent.actor_id()).is_some());
+    assert!(members.contains_key(&ta3.agent.actor_id()));
 
     let members = ta3.agent.members().read().states.clone();
     assert_eq!(members.len(), 1);
-    assert!(members.get(&ta1.agent.actor_id()).is_some());
+    assert!(members.contains_key(&ta1.agent.actor_id()));
 
     let members = ta2.agent.members().read().states.clone();
     assert_eq!(members.len(), 1);
-    assert!(members.get(&ta4.agent.actor_id()).is_some());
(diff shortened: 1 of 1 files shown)


crates/corro-admin/src/lib.rs
```diff
@@ -347,9 +347,20 @@ async fn handle_conn(
                     }
                     send_success(&mut stream).await;
                 }
-                Command::Cluster(ClusterCommand::MembershipStates) => {
+                Command::Cluster(ClusterCommand::MembershipStates { local }) => {
                     info_log(&mut stream, "gathering membership state").await;
 
+                    if local {
```
gorbak25: Ok so generally don't dump all peers we how about but only the ones we are connected to directly.

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -301,7 +302,7 @@ pub async fn handle_notifications(
                 info!("Member Up {actor:?} (result: {member_added_res:?})");
 
                 match member_added_res {
-                    MemberAddedResult::NewMember => {
+                    MemberAddedResult::NewMember | MemberAddedResult::Removed => {
                         debug!("Member Added {actor:?}");
                         counter!("corro.gossip.member.added", "id" => actor.id().0.to_string(), "addr" => actor.addr().to_string()).increment(1);
```
gorbak25: Do we want to include removed peers in that metric?

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -221,6 +221,7 @@ pub fn spawn_swim_announcer(agent: &Agent, gossip_addr: SocketAddr, mut tripwire
                     _ = timer.as_mut() => {}
                 }
 
+                // TODO: find way to find and filter out addrs with a different membership id
```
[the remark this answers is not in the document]
somtochiama: we keep sending out Announce messages even to peers that we learn have a different membership id. We could pass this information around, and exclude the peer's ip when we want to announce. But it is mostly an optimization so I left it out of this pr

PR #376, opened 2025-10-15
Fix restore script in antithesis tests

somtochiama:
- The restore script now tries if the container is shutting down if restarting corrosion after the restore fails. 
- update the debug script so it also accepts an antithesis moment and parse the required parameters for the mvd debugger.
- adds more details to some assertions


gorbak25 (approved):
LGTM!

PR #385, opened 2025-10-24
Use unnest for bulk operations in corrosion

gorbak25:
This PR uses #383 for efficient bulk operations on the `crsql_changes`, `__corro_buffered_changes`, `__corro_members` and bookie tables. To make writing array queries easier I've added the `unnest_param` which allows to use unnest in queries without too much boilerplate.

I've benchmarked the effect unnest had on full changes:
image
Previously in #377 (WARNING: The baseline chart has the red line on 50, now we use 100) the throughput values were platoing at 50k changes/s. With unnest that value went up to 70k changes/s with no plato in sight(probably perf will increase for even higher batch sizes). For the batch size we use perf went up slightly.

Larger changes with regards to buffered changes will be done in an separate PR.

### DIFF
--- a/crates/corro-types/src/agent.rs
+++ b/crates/corro-types/src/agent.rs
@@ -14,7 +14,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use antithesis_sdk::{assert_always, assert_unreachable};
+use antithesis_sdk::assert_unreachable;
 use arc_swap::ArcSwap;
 use camino::Utf8PathBuf;
 use compact_str::{CompactString, ToCompactString};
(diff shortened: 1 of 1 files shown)


crates/corro-agent/src/agent/util.rs
```diff
@@ -1271,57 +1279,109 @@ pub fn process_complete_version<T: Deref<Target = rusqlite::Connection> + Commit
...
+                    Option<String>, // comment
+                )>,
+            >>()?;
+        let details = json!({
+            "changes_len": len,
+            "returned_rowids_len": last_rowids.len(),
+            "query_plan": query_plan,
+            "vdbe_program": vdbe_program,
+        });
+        error!("Possible returning statement BUG: {details}");
```
somtochiama: Would we want to return an error here so we rollback the savepoint?

somtochiama (approved):
LGTM!

PR #388, opened 2025-10-30
Metrics for crsqlite storage overhead

gorbak25:
Analyses the table btree size every 15 minutes.


gorbak25:
Not worth it anymore. We got more efficient metrics now.

PR #402, opened 2025-11-27
Parse arrays with text format in postgres

somtochiama:


crates/corro-pg/src/codec.rs
```diff
@@ -169,3 +171,102 @@ where
...
+    let mut current = String::new();
+    let mut in_quotes = false;
+    let mut escape_next = false;
+    let mut depth = 0; // For nested arrays
+
+    for ch in input.chars() {
+        match ch {
+            '\\' if !escape_next => {
+                escape_next = true;
+            }
```
somtochiama: Possible bug in pgwire's implementation where we add escape baskslashes to the string was fixed here so it works fine with go's postgres client. I have opened an [issue](https://github.com/sunng87/pgwire/issues/357) in pgwire

crates/corro-pg/tests/tests.rs
```diff
@@ -925,37 +961,51 @@ async fn test_unnest_vtab() {
...
-                .query(
-                    "SELECT CAST(value0 AS text) FROM unnest(CAST($1 AS text[]))",
-                    &[&col1],
-                )
-                .await
-                .unwrap();
-            for (i, row) in rows.iter().enumerate() {
-                let val: String = row.get(0);
-                assert_eq!(val, col1[i]);
+            for format in [Format::Text, Format::Binary] {
```
gorbak25:
```suggestion
            for format in [Format::Text, Format::Binary] {
                  for func in ["unnest", "corro_unnest"] {
```

gorbak25 (approved):
lgtm 🚀

Issue #415, opened 2026-01-29
License unclear


Jake-Shadle:
We use a tool to enforce that all dependencies have licenses that we deem acceptable, which fails with a majority of the crates in this repository as most crates are missing the `license` or `license-file` fields in their cargo manifest. I was going to just open a PR to set the workspace license field and use that in all crates, but then noticed that there are _some_ crates (`consul-client`, `corro-api-types`, `corro-base-types`, `corro-client`, `corro-types`, and `tripwire`) that use `MIT`...however the `MIT` license file is not present in the repository, but the `Apache-2.0` is. The Apache license was added 3 years ago and hasn't been touched or used since AFAICT, so presumably the `MIT` license is the one to use, but again, kind of unclear since licensing isn't mentioned in the README as is typical.

PR #416, opened 2026-01-29
Use Apache-2.0 license for workspace

Jake-Shadle:
As mentioned in #415 most of the crates in this workspace did not specify their `license`, but the ones that did used `MIT`, so this sets ~`MIT`~`Apache-2.0` as the `license` in the workspace, along with the `edition`, `repository`, and `homepage` fields and uses them for all crates in the workspace.

This includes an update to the 2024 edition since the version in rust-toolchain is already specified as 1.88.0 and the 2024 edition was introduced in 1.85.0, but I can split out that change if wanted, but it's mostly just changes in rustfmt, only a few changes in the code itself.

Resolves: #415

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,8 +6,8 @@ members = [ ".antithesis/client/rust-load-generator",
 resolver = "2"
 
 [workspace.package]
-license = "MIT"
-edition = "2024"
+license = "Apache-2.0"
+edition = "2021"
 repository = "https://github.com/superfly/corrosion"
 homepage = "https://superfly.github.io/corrosion/"
 


crates/corro-agent/src/api/public/mod.rs
```diff
@@ -223,9 +223,7 @@ pub async fn api_v1_transactions(
                     Err(e) => Err(e),
                 }
             })
-            .collect::<Result<Vec<ExecResult>, ChangeError>>();
-
-        results
+            .collect::<Result<Vec<ExecResult>, ChangeError>>()
```
Jake-Shadle: This was just fixing a lint due to immediately returning a variable.

crates/corro-api-types/src/lib.rs
```diff
@@ -490,7 +490,7 @@ impl SqliteValue {
     }
 
     pub fn as_str(&self) -> Option<&str> {
-        if let Self::Text(ref s) = self {
```
Jake-Shadle: The `ref` here was completely pointless and triggered an error in the 2024 edition.

gorbak25 (approved):
lgtm

PR #431, opened 2026-02-27
Burstable change processing

gorbak25:
This PR adjust the amount of changes we try to process at once in real time based on the backlog of changes to sync.
The batch size increases exponentially with the amount of pending changes to process. In the default config: 
```
changes -> batch_size
0-499 -> 100
500-999 -> 500
1000-1999 -> 1000
2000-3999 -> 2000
...
16000-Inf -> 16000
```
This scaling can be configured via `apply_queue_min_batch_size`, `apply_queue_max_batch_size`, `apply_queue_step_base` in the configuration file.

This PR allows for processing of changes to be started early before accumulating a full batch. The decision for that is made by the `apply_queue_batch_threshold_ratio` parameter which sets a ratio(0-1) of how full a batch needs to be to start processing it. By default that value is set to `0.9` For ex. with batch size 100 and 90 pending changes when we finish processing a batch we would instantly spawn another batch instead of waiting for changes to accumulate.

The current batch size is available as the `corro.agent.changes.batch_size` metric.

### DIFF
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -18,7 +18,7 @@ use crate::{
     api::peer::parallel_sync,
     transport::Transport,
 };
-use antithesis_sdk::assert_sometimes;
+use antithesis_sdk::{assert_always, assert_sometimes};
 use camino::Utf8Path;
 use corro_types::{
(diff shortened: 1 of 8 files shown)


somtochiama (approved):
LGTM!

PR #435, opened 2026-03-03
Log in-progress queries when wal truncation fails

somtochiama:


crates/corro-types/src/sqlite.rs
```diff
@@ -120,34 +123,72 @@ fn tracing_callback_rw(ev: rusqlite::trace::TraceEvent) {
...
+            // skip trigger for subprograms.
+            if sql.starts_with("--") {
+                return;
+            }
+            let start_time = Instant::now();
+            // expanded sql so we differentiate similar queries ??
+            let sql = stmt_ref.expanded_sql();
+            if let Some(sql) = sql {
+                let queries_mu = IN_FLIGHT_QUERIES.get_or_default();
+                let mut inflight_queries = queries_mu.lock();
```
somtochiama: Not sure about this tbh, we are now doing inserts/removes at the start/end of every query and that might cause some perf hit.

PR #455, opened 2026-03-31
Reduce metrics cardinality, bookie consistency check, antithesis assertions

gorbak25:
- Introduces a new admin command (CheckBookieConsistency) that rebuilds bookie state from DB via BookedVersions::load_all_from_conn and compares it against in-memory bookie state to detect consistency drift.
- Runs the corrosion healtcheck on antithesis. This should catch sync performance degradation in antithesis.
- Changes corro.db.gaps.sum to bucket by peer_state instead of actor_id to avoid high-cardinality series.
- Introduces corro.db.buffered.changes.v2.total and corro.db.buffered.changes.v2.oldest_age_seconds for buffered-change visibility with bounded cardinality. The metric got renamed so the old metrics don't slow down grafana.
- Removes high-cardinality transport per-address metrics; keeps delta tracking and logs only increases for cwnd, congestion_events, and black_holes_detected.
- Buckets transport/connect/send metrics by traffic class (sync, broadcast, foca) and wires traffic tagging through open_bi, send_uni, and send_datagram. This way we can distinguish if sync or broadcast uses up more traffic. This tagging relies on the assumption that:
    * `open_bi` is used primarily in sync
    * `send_datagram` by FOCA/SWIM
    * `send_uni` by broadcast traffic
- Adds v2 to the name of some metrics to distinguish the old high cardinality series from the new low cardinality ones
- Introduces `corro.transport.rtt.v2.seconds` which is an histogram of the rtt of peers

### DIFF
--- a/crates/corro-admin/src/lib.rs
+++ b/crates/corro-admin/src/lib.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     fmt::Display,
     time::{Duration, Instant},
 };
@@ -6,7 +7,7 @@ use std::{
 use camino::Utf8PathBuf;
 use corro_types::{
     actor::{ActorId, ClusterId},
-    agent::{Agent, Bookie},
+    agent::{Agent, BookedVersions, Bookie},
     base::{CrsqlDbVersion, CrsqlSeq},
     broadcast::{FocaCmd, FocaInput},
     sqlite::SqlitePoolError,
@@ -113,6 +114,7 @@ pub enum Command {
 pub enum SyncCommand {
     Generate,
     ReconcileGaps,
+    CheckBookieConsistency,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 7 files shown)


crates/corro-admin/src/lib.rs
```diff
@@ -707,3 +757,66 @@ fn collapse_gaps(
...
+fn check_bookie_consistency(
+    conn: &rusqlite::Connection,
+    bookie: &Bookie,
+) -> rusqlite::Result<BookieConsistencyReport> {
+    let _bookie_write_guard = bookie.write_lock_blocking();
+
+    let db_map = BookedVersions::load_all_from_conn(conn)?;
+
+    let in_memory_map: HashMap<ActorId, BookedVersions> = {
+        let guard = bookie.owned_guard();
```
somtochiama, line 783: do we want to drop the write conn + bookie lock once we've created the two maps?

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -1065,13 +1072,20 @@ pub async fn handle_sync(
...
+        .sum::<u64>();
+    gauge!("corro.sync.client.needed.v2", "traffic" => "sync").set(needed_total as f64);
+
+    let head_max = sync_state
+        .heads
+        .values()
+        .map(|version| version.0)
+        .max()
+        .unwrap_or_default();
+    gauge!("corro.sync.client.head.v2", "traffic" => "sync").set(head_max as f64);
```
somtochiama: Is this metric still useful if it is no longer per actor? maybe we can remove it completely?

crates/corro-admin/src/lib.rs
gorbak25: Good point!

somtochiama (approved):
LGTM @gorbak25 !

crates/corro-agent/src/agent/handlers.rs
gorbak25: Good catch!

PR #459, opened 2026-04-14
Implement max_mtu validation and builder method

Ritish134:
Added validation for max_mtu in Config and updated builder to set max_mtu and disable GSO


gorbak25 (approved):
Thank you for the contribution!

PR #460, opened 2026-04-15
Make gossip max packet size configurable

Ritish134:
Foca uses max_packet_size to size its SWIM gossip UDP datagrams. 1178 is a conservative hardcoded value. The problem is if `max_mtu` = 1420 is set, foca still limits its packets to 1178

#### The calculation
```
max_packet_size = max_mtu - IP header - UDP header - QUIC short header
               = max_mtu - 40 (IPv6) - 8 (UDP) - 13 (QUIC short header)
               = max_mtu - 61
```
Using `64 bytes` as a conservative overhead to give a small safety margin.

### DIFF
--- a/crates/corro-agent/src/broadcast/mod.rs
+++ b/crates/corro-agent/src/broadcast/mod.rs
@@ -2,7 +2,7 @@ use std::{
     cmp,
     collections::{hash_map::Entry, HashMap, HashSet, VecDeque},
     net::SocketAddr,
-    num::NonZeroU32,
+    num::{NonZeroU32, NonZeroUsize},
     pin::Pin,
     sync::{
         atomic::{AtomicU32, Ordering},
@@ -136,7 +136,7 @@ pub fn runtime_loop(
     let max_mtu = agent.config().gossip.max_mtu;
     let config = Arc::new(RwLock::new(make_foca_config(
         1.try_into().unwrap(),
-        max_mtu,
+        agent.config().gossip.max_mtu,
     )));
 
     let mut foca = Foca::with_custom_broadcast(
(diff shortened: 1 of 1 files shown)


crates/corro-agent/src/broadcast/mod.rs
```diff
@@ -1016,13 +1020,19 @@ fn foca_state_is_active(state: &foca::State) -> bool {
...
     config.remove_down_after = Duration::from_secs(2 * 24 * 3600);
 
-    // max payload size for udp datagrams, use a safe value here...
-    // TODO: calculate from smallest max datagram size for all QUIC conns
-    config.max_packet_size = 1178.try_into().unwrap();
+    // Derive max packet size from configured MTU if available.
+    // Overhead: 40 bytes IPv6 + 8 bytes UDP + ~13 bytes QUIC short header,
+    // rounded up to 64 for safety. Falls back to a conservative 1178 if
+    // no MTU is configured
+    config.max_packet_size = max_mtu
```
gorbak25, line 1029: The `max_mtu` in the gossip config is slightly misleading - it's the effective [max udp payload size](https://github.com/superfly/corrosion/blob/1b0fcf1b12a81c5394dd4b9e5a7815b0ab95d575/doc/config/gossip.md?plain=1#L59), not the link size. For link MTU 1420 and IPv6 u would set it to 1372, while on IPv4 u would use 1392.

So please only subtract the size of the QUIC header here.

gorbak25 (changes requested):

crates/corro-agent/src/broadcast/mod.rs
Ritish134: done

gorbak25 (approved):
Looks good! Thank you ❤️
Will smoke test it and merge

PR #472, opened 2026-05-06
Record when corro channel goes above threshold capacity 

somtochiama:
Previously we skipped recording when capacity was above the threshold if fewer than 30 ticks had passed since the last record. This caused some recovereis from low capacity to be missed, leading to inaccurate metrics and false alerts. This PR fixes it by recording if previous capacity was below threshold too.

### DIFF
--- a/crates/corro-types/src/channel.rs
+++ b/crates/corro-types/src/channel.rs
@@ -78,18 +78,22 @@ pub fn bounded<T: Send + 'static>(
         let mut ticks_since_report = 0;
         let mut tick = interval(Duration::from_secs(1));
         tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        let mut prev = inner_channel.capacity();
         loop {
             tick.tick().await;
             if inner_channel.is_closed() {
                 break;
             }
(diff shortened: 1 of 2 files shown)


crates/corro-types/src/channel.rs
```diff
@@ -78,18 +78,22 @@ pub fn bounded<T: Send + 'static>(
...
         tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        let mut prev = inner_channel.capacity();
         loop {
             tick.tick().await;
             if inner_channel.is_closed() {
                 break;
             }
             let current = inner_channel.capacity();
-            if current < threshold || ticks_since_report >= 30 {
+            if prev != current
```
gorbak25: if prev == current remains for more than the gauge GC period then the metric will just dissapear from grafana

somtochiama: good catch. I've changed the statement so we record every 30 ticks.

gorbak25 (commented):
lgtm!

gorbak25 (approved):
lgtm

PR #475, opened 2026-05-11
Prepare for v1.0.0 release

somtochiama:
This pull request:
- updates changelog for v1.0.0
- add readme and bump version for public crates
- improve documentation for public crates
- adds workflow for publishing public crates from CI


crates/corro-client/src/lib.rs
```diff
@@ -716,6 +716,87 @@ pub enum Error {
     ExpectedQueryId,
 }
 
+pub async fn read_files_from_paths<P: AsRef<Path>>(
```
gorbak25: How does this relate to #476

gorbak25 (approved):
lgtm

crates/corro-client/src/lib.rs
somtochiama: It will get removed.

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+, or bad / unwanted data is spreading through the cluster and restoring a known-good
+snapshot is faster than trying to delete or repair data. We call this restore-from-snapshot
+workflow is called a reseed; the rest of this document outlines how to do it safely.
+
+```admonish warning
+Reseeding the cluster from a snapshot means that any writes accepted
+after the snapshot was taken will be lost. Either stop all writes before
+taking the snapshot, or be prepared to replay the missing writes after the
+migration (see [Re-insert any missing data](#re-insert-any-missing-data)).
+```
```
gorbak25:
```suggestion
You may need to restart every Corrosion node from an earlier snapshot. Common reasons include:
 
- **Major corrosion updates** — a version upgrade might have changed the database structure of internal tables in an incompatible way.
- **Breaking schema change** — you need to make a breaking change to your database schema that cr-sqlite forbids like for ex. removing a column.
- **Bad data spreading** — unwanted data is propagating through the cluster, and restoring a known-good snapshot is faster than deleting or repairing it.
We call this restore-from-snapshot workflow a **reseed**. The rest of this document explains how to do it safely.
 
> **Warning — data loss is expected.**
> Reseeding from a snapshot means any writes accepted *after* the snapshot was taken will be lost. Either stop all writes before taking the snapshot, or be prepared to replay the missing writes afterward (see [Re-insert any missing data](#3-re-insert-any-missing-data)).

```

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Overview
+
+The migration is performed in three phases:
+
+1. On a single old node, prepare a clean snapshot suitable (Skip this if you already have a snapshot).
```
gorbak25:
```suggestion
1. On a single old node, create a clean snapshot (skip if you already have one).
```

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+
+Pick one node in the old cluster and use it as the source of truth for the
+snapshot. All other nodes will be restored from the backup created from this node.
+
+### 1.1 Create a backup
+
+While the corrosion agent is still running on the source node:
+
+```bash
+corrosion backup /path/to/prev-snapshot.db
```
gorbak25:
```suggestion
corrosion backup /path/to/snapshot.db
```

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+so the backup is often smaller than the running database.
+
+### 1.2 Bump the cluster id
+
+A new `cluster_id` ensures nodes that come up with the new snapshot don't
+accidentally talk to any nodes still on the old snapshot. Run this against
+the backup you just created:
+
+```bash
+sqlite3 /path/to/old-snapshot.db <<'SQL'
```
gorbak25:
```suggestion
sqlite3 /path/to/snapshot.db <<'SQL'
```

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+corrosion backup /path/to/prev-snapshot.db
+```
+
+`corrosion backup` runs `VACUUM INTO` and strips per-node state
+(`crsql_site_id`, `__corro_members`, `__corro_subs`, consul hash tables),
+so the backup is often smaller than the running database.
+
+### 1.2 Bump the cluster id
+
+A new `cluster_id` ensures nodes that come up with the new snapshot don't
```
gorbak25, line 40: Makes me thing why isn't this an option for corrosion backup

somtochiama: heh, good point. it should probably be to reduce manual queries

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+DO UPDATE SET value = value + 1;
+SQL
+```
+
+If no `cluster_id` is set (it defaults to `0` when unset) this inserts `1`;
+otherwise it increments the existing value by one.
+
+### 1.3 Drop cr-sqlite internal tables (optional)
+
+If this is due to version upgrade and the schema of the cr-sqlite internal or your database is getting big
```
gorbak25:
```suggestion
Do this if either of the following applies:
 
- You're upgrading versions and the cr-sqlite internal table schema has changed.
- Your database has grown large and you want a clean reset.
Dropping these tables lets them be recreated under the new schema on startup. Run everything below against the **snapshot file**, not the live database.

**Drop the clock and primary-key tables:**
 
```bash
sqlite3 snapshot.db ".mode list" \
  "SELECT 'DROP TABLE ' || name || ';' \
   FROM sqlite_schema \
   WHERE type = 'table' \
     AND (name LIKE '%__crsql_clock' \
       OR name LIKE '%__crsql_pks');" \
  | sqlite3 snapshot.db
```
 
This generates `DROP TABLE` statements from the schema, then pipes them back into `sqlite3` to execute.
 
**Drop bookkeeping and reclaim space:**
 
```sql
sqlite3 /path/to/snapshot.db <<'SQL'
DROP TABLE IF EXISTS __corro_bookkeeping;
VACUUM;
SQL
```
 
The `VACUUM` reclaims space freed by the dropped tables and keeps the snapshot small for transfer.
 

```

File: doc/reseeding.md [hunk 8]
```diff
@@ -0,0 +1,126 @@
...
+Compress the snapshot and upload it to a location every node can reach (S3, an internal artifact store, etc.):
+
+```bash
+pigz /path/to/old-snapshot.db
+# upload /path/to/old-snapshot.db.gz somewhere
+```
+
+## 2. Restore every node from the snapshot
+
+For each node in the cluster:
```
gorbak25:
```suggestion
Repeat these steps on **every node** in the cluster:
```

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+
+1. (if this is a version upgrade) Install the new `corrosion` binary .
```
gorbak25:
```suggestion
1. **(Version upgrade only)** Install the new `corrosion` binary.
```

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+2. Download and decompress the snapshot to a local path.
+3. Stop the Corrosion agent. `corrosion restore` refuses to run while the agent is up.
+4. Restore it:
```
gorbak25:
```suggestion
4. Restore:
```

doc/reseeding.md
```diff
@@ -0,0 +1,126 @@
...
+
+   ```bash
+   corrosion restore /path/to/old-snapshot.db
+   ```
+
+5. Start the v1 agent.
```
gorbak25:
```suggestion
5. Start the Corrosion agent.
 ```

File: doc/reseeding.md [hunk 12]
```diff
@@ -0,0 +1,126 @@
...
+
+Once a v1 node starts, it will create the `<table>__crsql_clock` and
```
gorbak25:
```suggestion
Once a node starts, it will create the `<table>__crsql_clock` and
```

gorbak25 (approved):
looks good

PR #476, opened 2026-05-13
Remove v1/migration endpoint

somtochiama:
Previously the `v1/migration` accepted statements to update corrosion's schema, however that means that the in-memory schema can diverge from what is defined on disk (via `schema_paths` in config)

This pull request deprecates the endpoint and updates the `corrosion reload` command to send an admin command instead of calling the migration endpoint. It also removes a bunch of places where the schema is reloaded twice.


crates/corro-agent/src/agent/util.rs
```diff
@@ -1428,6 +1433,70 @@ pub fn process_complete_version<T: Deref<Target = rusqlite::Connection> + Commit
     Ok::<_, rusqlite::Error>((known_version, new_changeset, changes_per_table))
 }
 
+pub async fn execute_schema_from_paths(agent: &Agent) -> eyre::Result<()> {
```
somtochiama, line 1418: file is getting too long. but i'll split things up in a different PR

gorbak25 (commented):
Can we have some tests for `execute_schema_from_paths` via some sql files in /tmp ?

PR #477, opened 2026-05-13
Bound bi-stream handler concurrency at accept site

MorganaFuture:
A peer could open unbounded bi-streams and exhaust memory before the inner `agent.limits().sync` check inside `serve_sync` ever fired. Move the try_acquire up to `spawn_bipayload_handler`, before `tokio::spawn`, and write `MaxConcurrencyReached` on the stream when no permits remain.

The duplicate check inside `serve_sync` is removed: keeping both would consume two permits per sync from the same 3-permit pool.

### DIFF
--- a/crates/corro-agent/src/agent/bi.rs
+++ b/crates/corro-agent/src/agent/bi.rs
@@ -1,7 +1,9 @@
-use crate::api::peer::serve_sync;
+use crate::api::peer::{encode_write_sync_msg, serve_sync};
+use bytes::BytesMut;
 use corro_types::{
     agent::{Agent, Bookie},
     broadcast::{BiPayload, BiPayloadV1},
+    sync::{SyncMessage, SyncMessageV1, SyncRejectionV1},
 };
 use metrics::counter;
 use speedy::Readable;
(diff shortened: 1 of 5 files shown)


crates/corro-agent/src/agent/bi.rs
```diff
@@ -51,11 +53,61 @@ pub fn spawn_bipayload_handler(
                 conn.remote_address()
             );
 
-            // TODO: implement concurrency limit for sync requests
+            // Bound the number of in-flight bi-handler tasks. All bi-streams
+            // are sync requests today, so we reuse `Limits.sync` rather than
+            // adding a parallel counter. Acquiring before `tokio::spawn`
+            // prevents an unbounded peer from exhausting tasks/memory just
+            // by opening streams.
+            let permit = match agent.limits().sync.clone().try_acquire_owned() {
```
somtochiama, line 61: Maybe we'd want to make this configurable? Right now we have it set to three and there's no way to tune it.

MorganaFuture: Done

PR #492, opened 2026-06-24
Introduce plumtree broadcast method in corrosion.

somtochiama:
This pull requests adds a new plumtree protocol for broadcasting changes across a cluster. It is modelled after the original plumtree protocol outlined in [the original paper](https://www.dpss.inesc-id.pt/~ler/reports/srds07.pdf) but it differs in a number of key ways.
1. Full membership: Corrosion's plumtree protocol is built top of SWIM, a full membership protocol while the paper uses hyparview which is a partial membership. A node knows about every node in the cluster, and randomly selects eager and lazy peers from known peers.
2. Multi-sender: The original plumtree protocol is ideal for single-sender trees but we have multi-master setup where every node can disseminate changes. To make the tree stable under many concurrent senders:
     - We tolerate a configurable number of duplicates: a PRUNE is only sent once a duplicate count crosses prune_threshold, rather than on the first duplicate. We trade a little duplication for tree stability, since we're trying to find peers that are good sources across multiple senders. This also has the additional benefit that each node has multiple links for a particular sender so that a failed node doesn't 
     - Links are asymmetric: receiving a gossip from a peer does not add them to our eager set. A peer is only promoted to eager when it deliberately sends a GRAFT.
4. Fanout: We use a much wider fanout than recommended in the paper, while the paper recommends a fanout of five we adjust based on the current number of nodes in the cluster to (log N * 3). This is mostly because we are aiming for lower latency, and can tolerate some duplication.
5. RTT-aware peer selection: Rather than picking eager and lazy peers at random, we bias toward low-RTT peers for the eager set, with a handful of far ones mixed in so messages still spread quickly across clusters that aren't close to each other. This would still get adjusted via prunes and grafts by the protocol.

The plumtree method can be enabled by adding this to the node config
```
[gossip]
broadcast = { plumtree = {} }
```

This was partly inspired by: https://github.com/johnnywale/memberlist-plumtree

### DIFF
--- a/crates/corro-agent/src/plumtree/mod.rs
+++ b/crates/corro-agent/src/plumtree/mod.rs
@@ -27,7 +27,11 @@ use metrics::{counter, gauge, histogram};
 use plum_foca::{Payload, PlumPrio, PlumtreeState, RttInfo, SeenStore, Timer};
 use rangemap::RangeInclusiveSet;
 use speedy::Writable;
-use tokio::{sync::mpsc, task::JoinSet, time::interval};
+use tokio::{
+    sync::mpsc,
+    task::JoinSet,
+    time::{interval, MissedTickBehavior},
+};
 use tokio_util::codec::{Encoder, LengthDelimitedCodec};
 use tracing::{error, info, trace, warn};
 use tripwire::Tripwire;
@@ -35,15 +39,13 @@ use tripwire::Tripwire;
 use crate::{
     agent::util::log_at_pow_10,
     broadcast::{try_transmit_uni, TimerSpawner, TransmitError, TransmitRateLimiter},
-    transport::TransportExt,
+    transport::Transport,
 };
 
 #[derive(Debug)]
 struct SeenEntry {
     seqs: Option<RangeInclusiveSet<CrsqlSeq>>,
     last_seq: Option<CrsqlSeq>,
-    #[allow(unused)]
-    round: plum_foca::Round,
     duplicate_count: u32,
 }
 
@@ -79,7 +81,7 @@ impl SeenStore<ChangeId> for ChangeSeenStore {
         self.entries.len()
     }
 
-    fn observe(&mut self, id: ChangeId, round: plum_foca::Round) -> Option<u32> {
+    fn observe(&mut self, id: ChangeId, _round: plum_foca::Round) -> Option<u32> {
         // Already applied via sync/apply or entry was dropped
         let already_booked = !self.has_cache_entry(&id) && self.contains_booked(&id);
 
@@ -95,7 +97,6 @@ impl SeenStore<ChangeId> for ChangeSeenStore {
                     e.insert(SeenEntry {
                         seqs: Some(incoming),
                         last_seq: Some(*last_seq),
-                        round,
                         duplicate_count: 0,
                     });
 
@@ -120,6 +121,7 @@ impl SeenStore<ChangeId> for ChangeSeenStore {
                         // if we seen this partial change, but we are still incomplete, return zero duplicates.
                         // We don't want partials to cause a prune while we are in the middle of receiving the change.
                         if !is_complete {
+                            counter!("corro.plumtree.partial.duplicate").increment(1);
                             return Some(0);
                         }
 
@@ -146,7 +148,6 @@ impl SeenStore<ChangeId> for ChangeSeenStore {
                             SeenEntry {
                                 seqs: None,
                                 last_seq: None,
-                                round,
                                 duplicate_count: 0,
                             },
                         );
@@ -338,9 +339,9 @@ impl plum_foca::Runtime<ChangeId, PlumtreePayload, ActorId> for CorrosionPlumtre
     }
 }
 
-pub async fn spawn_plumtree_loop<T: TransportExt + Clone + Send + 'static>(
+pub async fn spawn_plumtree_loop(
     agent: Agent,
-    transport: T,
+    transport: Transport,
     rx_plumtree: CorroReceiver<PlumtreeInput>,
     rx_plumtree_updates: CorroReceiver<PlumtreeUpdates>,
     tx_changes: CorroSender<(ChangeV1, ChangeSource, Option<BroadcastV1>)>,
@@ -353,15 +354,17 @@ pub async fn spawn_plumtree_loop<T: TransportExt + Clone + Send + 'static>(
         .cloned()
         .unwrap_or_default();
 
+    let max_queue_len = agent.config().perf.processing_queue_len;
+
     let config = plum_foca::Config {
         ihave_timeout: Duration::from_millis(150),
         optimization_threshold: plumtree_config.optimization_threshold,
-        max_cached_payloads: 5000,
+        max_cached_payloads: max_queue_len,
         num_eager: None,
         min_lazy: None,
         max_lazy: None,
         prune_threshold: plumtree_config.prune_threshold,
-        max_received_entries: 10000,
+        max_received_entries: max_queue_len,
         prune_throttle: plumtree_config.prune_throttle_secs.map(Duration::from_secs),
         eager_ratios: plumtree_config.eager_ratios,
     };
@@ -378,9 +381,9 @@ pub async fn spawn_plumtree_loop<T: TransportExt + Clone + Send + 'static>(
     .await;
 }
 
-pub async fn plumtree_loop<T: TransportExt + Clone + Send + 'static>(
+pub async fn plumtree_loop(
     agent: Agent,
-    transport: T,
+    transport: Transport,
     mut rx_plumtree: CorroReceiver<PlumtreeInput>,
     mut rx_plumtree_updates: CorroReceiver<PlumtreeUpdates>,
     tx_changes: CorroSender<(ChangeV1, ChangeSource, Option<BroadcastV1>)>,
@@ -572,20 +575,14 @@ pub async fn plumtree_loop<T: TransportExt + Clone + Send + 'static>(
     }
 }
 
-#[derive(Debug)]
 struct PendingPlumtreeSend {
     peers: Vec<ActorId>,
-    payload: Bytes,
+    payload: BytesMut,
 }
 
-struct P1GossipBatch {
-    peers: Vec<ActorId>,
-    buf: BytesMut,
-}
-
-async fn send_messages_loop<T: TransportExt + Clone + Send + 'static>(
+async fn send_messages_loop(
     agent: Agent,
-    transport: T,
+    transport: Transport,
     mut rx_msgs: CorroReceiver<(PlumPrio, Vec<ActorId>, PlumtreeMsgV1)>,
 ) {
     const MAX_INFLIGHT: usize = 500;
@@ -609,9 +606,9 @@ async fn send_messages_loop<T: TransportExt + Clone + Send + 'static>(
 
     let mut p0_queue: VecDeque<PendingPlumtreeSend> = VecDeque::new();
     let mut p1_queue: VecDeque<PendingPlumtreeSend> = VecDeque::new();
-    let mut p1_gossip_batch = P1GossipBatch {
+    let mut p1_gossip_batch = PendingPlumtreeSend {
         peers: Vec::new(),
-        buf: BytesMut::new(),
+        payload: BytesMut::new(),
     };
     let mut gossip_batch_interval = interval(P1_GOSSIP_BATCH_INTERVAL);
     let mut join_set = JoinSet::new();
@@ -623,70 +620,58 @@ async fn send_messages_loop<T: TransportExt + Clone + Send + 'static>(
     }))
     .with_middleware();
 
-    #[allow(clippy::large_enum_variant)]
-    enum Branch {
-        Msg((PlumPrio, Vec<ActorId>, PlumtreeMsgV1)),
-        GossipBatchDeadline,
-    }
-
     loop {
-        let branch = tokio::select! {
+        let msg = tokio::select! {
             biased;
             _ = join_set.join_next(), if !join_set.is_empty() => {
                 continue;
             },
+            _ = gossip_batch_interval.tick(), if batch_gossip => {
+                if !p1_gossip_batch.payload.is_empty() {
+                    p1_queue.push_back(PendingPlumtreeSend {
+                        peers: std::mem::take(&mut p1_gossip_batch.peers),
+                        payload: std::mem::take(&mut p1_gossip_batch.payload),
+                    });
+                }
+                continue;
+            },
             msg = rx_msgs.recv() => match msg {
-                Some(msg) => Branch::Msg(msg),
+                Some(msg) => msg,
                 None => {
                     warn!("plumtree send loop: message channel closed");
                     break;
                 }
             },
-            _ = gossip_batch_interval.tick() => Branch::GossipBatchDeadline,
         };
 
         let mut rate_limited = false;
+        let (prio, peers, msg) = msg;
+        trace!("plumtree: msg: {msg:?}, peers: {peers:?}");
+        let p1_gossip = matches!(&msg, PlumtreeMsgV1::Gossip(_));
+        let payload =
+            match encode_plumtree_wire(cluster_id, &mut codec, &mut ser_buf, &mut frame_buf, msg) {
+                Ok(payload) => payload,
+                Err(()) => continue,
+            };
 
-        match branch {
-            Branch::GossipBatchDeadline => {
-                if !p1_gossip_batch.buf.is_empty() {
-                    p1_queue.push_back(PendingPlumtreeSend {
-                        peers: std::mem::take(&mut p1_gossip_batch.peers),
-                        payload: p1_gossip_batch.buf.split().freeze(),
-                    });
-                }
+        if batch_gossip && p1_gossip {
+            // gossip is sent to latest eager peers
+            p1_gossip_batch.peers = peers;
+            p1_gossip_batch.payload.extend_from_slice(&payload);
+            if p1_gossip_batch.payload.len() >= P1_GOSSIP_BATCH_CUTOFF {
+                p1_queue.push_back(PendingPlumtreeSend {
+                    peers: std::mem::take(&mut p1_gossip_batch.peers),
+                    payload: std::mem::take(&mut p1_gossip_batch.payload),
+                });
             }
-            Branch::Msg((prio, peers, msg)) => {
-                trace!("plumtree: msg: {msg:?}, peers: {peers:?}");
-                let p1_gossip = matches!(&msg, PlumtreeMsgV1::Gossip(_));
-                let payload = match encode_plumtree_wire(
-                    cluster_id,
-                    &mut codec,
-                    &mut ser_buf,
-                    &mut frame_buf,
-                    msg,
-                ) {
-                    Ok(payload) => payload,
-                    Err(()) => continue,
-                };
-
-                if batch_gossip && p1_gossip {
-                    // gossip is sent to latest eager peers
-                    p1_gossip_batch.peers = peers;
-                    p1_gossip_batch.buf.extend_from_slice(&payload);
-                    if p1_gossip_batch.buf.len() >= P1_GOSSIP_BATCH_CUTOFF {
-                        p1_queue.push_back(PendingPlumtreeSend {
-                            peers: std::mem::take(&mut p1_gossip_batch.peers),
-                            payload: p1_gossip_batch.buf.split().freeze(),
-                        });
-                    }
-                } else {
-                    let pending = PendingPlumtreeSend { peers, payload };
-                    match prio {
-                        PlumPrio::P0 => p0_queue.push_back(pending),
-                        PlumPrio::P1 => p1_queue.push_back(pending),
-                    }
-                }
+        } else {
+            let pending = PendingPlumtreeSend {
+                peers,
+                payload: BytesMut::from(payload),
+            };
+            match prio {
+                PlumPrio::P0 => p0_queue.push_back(pending),
+                PlumPrio::P1 => p1_queue.push_back(pending),
             }
         }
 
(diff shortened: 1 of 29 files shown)


doc/config/gossip.md
```diff
@@ -62,6 +62,38 @@ This should be your "effective" MTU: `network interface's MTU - IP header size -
 
 Certain environments don't support GSO (Generic Segmentation Offload). This is detected by the QUIC implementation, but it's possible to pre-emptively disable it to avoid re-trying the initial packets without GSO as it is detected as unavailable.
 
+#### `gossip.broadcast`
+
+Selects the algorithm used to disseminate changes across the cluster. Two methods are available:
```
gorbak25: Please note that the entire cluster needs to use the same broadcast algorithm. U can't have part of the cluster use gossip and the rest plumtree.

doc/config/gossip.md
```diff
@@ -62,6 +62,38 @@ This should be your "effective" MTU: `network interface's MTU - IP header size -
...
+
+Higher values tolerate more duplication in exchange for a more stable tree; lower values prune more aggressively, reducing redundant payload traffic but causing more tree churn. Defaults to `5`.
+
+##### `optimization_threshold`
+
+Threshold, in gossip rounds, for switching to a shorter path. When a node learns via a lazy `IHave` that a peer could have delivered a message at least this many rounds sooner than the eager peer that actually delivered it, it grafts the shorter path and prunes the longer one.
+
+Lower values optimize the tree more aggressively (more `GRAFT`/`PRUNE` churn); higher values leave the tree alone unless the improvement is large. Defaults to `7`.
+
+##### `batch_gossip`
```
gorbak25, line 96: Does it seem that we will use that option?

crates/plum-foca/tests/sim.rs
```diff
@@ -0,0 +1,1472 @@
...
+        // time for a fully-delivered message to reach the whole cluster
+        let mut full: Vec<u64> = self
+            .msgs
+            .values()
+            .filter(|m| m.deliveries as u64 == self.params.n as u64 - 1)
+            .map(|m| m.last_delivery_at - m.sent_at)
+            .collect();
+        full.sort_unstable();
+
+        // RMR per message, in broadcast order, to show tree convergence
```
gorbak25: RMR? What's that? Is it smth bout redundancy? 🤔

crates/plum-foca/tests/sim.rs
```diff
@@ -0,0 +1,1472 @@
...
+    assert!(
+        report.rmr_first > 1.0,
+        "gossip RMR {:.2} unexpectedly low",
+        report.rmr_first
+    );
+}
+
+/// Head-to-head: the current foca gossip broadcast vs plumtree at 2k, on the
+/// same topology, printed as one comparison table. This is the "what does
+/// plumtree buy us" test. RMR is the key number (redundant transmissions per
```
gorbak25: Why not name it RTD? or unnecessary_transmissions?

crates/plum-foca/tests/sim.rs
```diff
@@ -0,0 +1,1472 @@
...
+
+#[derive(Default)]
+struct Outbox {
+    sent: Vec<(NId, PlumtreeMsg<MId, SimPayload, NId>)>,
+    scheduled: Vec<(Timer<MId, NId>, Duration)>,
+    delivered: Vec<SimPayload>,
+    notification_stats: NotificationStats,
+    /// current virtual time in µs, kept in sync with `Sim::now`
+    now_us: u64,
+}
```
gorbak25:
```suggestion
struct SimRuntime {
    // Collects outgoing protocol messages  
    // (DstNodeId, PlumtreeMsg<MessageId, .., SrcNodeId>)
    outbox: Vec<(NId, PlumtreeMsg<MId, SimPayload, NId>)>,
    timers: Vec<(Timer<MId, NId>, Duration)>,
    // Payloads the protocol delivered this step.
    inbox: Vec<SimPayload>,
    notification_stats: NotificationStats,
    /// current virtual time in µs, kept in sync with `Sim::now`
    now_us: u64,
}
```


Cause that's not simply a single node message outbox.

crates/plum-foca/tests/sim.rs
```diff
@@ -0,0 +1,1472 @@
...
+struct Sim {
+    params: Params,
+    now: u64, // virtual time, µs
+    next_seq: u64,
+    queue: BinaryHeap<Scheduled<Event>>,
+    states: Vec<PlumtreeState<MId, SimPayload, NId, SimSeenStore>>,
+    region_of: Vec<u8>,
+    rtt_ms: Vec<Vec<u64>>,
+    rng: StdRng,
+    outbox: Outbox,
```
gorbak25:
```suggestion
    runtime: SimRuntime,
```

crates/plum-foca/tests/sim.rs
```diff
@@ -0,0 +1,1472 @@
+//! Deterministic discrete-event simulation of the Plumtree protocol at
```
gorbak25, line 1: Should we add node failures(nodes getting down-up gracefully/nodes crashing ungracefully) ? Currently the only failure mode simulated here is message/packet loss(params.loss).

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+pub trait MessageId: Clone + Eq + Hash + Debug + Send + 'static {}
+
+pub trait Payload: Clone + Debug + Send + 'static {
+    type MessageId: MessageId;
+    type NodeId: NodeId;
+    fn message_id(&self) -> Self::MessageId;
+    fn origin(&self) -> Self::NodeId;
+}
+
+pub trait NodeId: Copy + Eq + Hash + Ord + Debug + Send + 'static {}
```
gorbak25: Do we need 'static here?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+
+impl<T> MessageId for T where T: Clone + Eq + Hash + Debug + Send + 'static {}
+impl<T> NodeId for T where T: Copy + Eq + Hash + Ord + Debug + Send + 'static {}
+
+pub type Round = u32;
```
gorbak25, line 30: How much time does it take for the round counter to overflow?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+}
+
+/// Tunable parameters for the Plumtree protocol.
+#[derive(Debug, Clone)]
+pub struct Config {
+    /// How long to wait after receiving an IHave before sending a GRAFT
+    /// if the full message hasn't arrived yet.
+    pub ihave_timeout: Duration,
+    /// Optimization threshold in rounds. Controls when a node receiving
+    /// a GOSSIP will attempt to GRAFT a shorter-path peer (via IHave)
```
gorbak25:
```suggestion
    /// Optimisation threshold in rounds. Controls the number of rounds a node can miss a message for before we will attempt to GRAFT a shorter-path peer (via IHave)
```

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+    pub min_lazy: Option<usize>,
+    /// Maximum number of lazy peers. `None` defaults to `num_eager * 2`.
+    pub max_lazy: Option<usize>,
+    /// Maximum number of times a message can be received before pruning sender.
+    /// We trade some duplication for tree stability.
+    pub prune_threshold: u32,
+    /// cap on seen entries; oldest are evicted when exceeded.
+    pub max_received_entries: usize,
+    /// cap on cached payload used to respond to GRAFT requests.
+    pub max_cached_payloads: usize,
```
gorbak25, line 129: Do we store full payloads or payload hashes?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+            Some(0 | 1) => Self::Near,
+            Some(2 | 3) => Self::Mid,
+            Some(_) | None => Self::Far,
+        }
+    }
+}
+
+/// Tunable parameters for the Plumtree protocol.
+#[derive(Debug, Clone)]
+pub struct Config {
```
gorbak25, line 109: I don't see an option for configuring for how long does a round last - this will also depend on the cluster.

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+
+/// Resolved eager/lazy peer targets used by the protocol.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct FanoutTargets {
+    num_eager: usize,
+    min_lazy: usize,
+    max_lazy: usize,
+}
+
+fn resolve_fanout(known_peers: usize, config: &Config) -> FanoutTargets {
```
gorbak25, line 161: Can the fanout change over time when the cluster grows or will corrosion need to get restarted?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+    fn deliver(&mut self, payload: P);
+
+    /// Request the caller to schedule a timer. When the duration elapses,
+    /// the caller must invoke `PlumtreeState::timer_fired`.
+    fn schedule(&mut self, timer: Timer<I, N>, after: Duration);
+
+    /// Observable protocol event.
+    fn notify(&mut self, notification: Notification<'_, I, N>);
+
+    /// Current time. The state machine is sans-IO and owns no clock, so the
```
gorbak25:
```suggestion
    /// Current time. The state machine is sans-IO(sans-io.readthedocs.io) and owns no clock, so the
```

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+pub enum PlumtreeMsg<I, P, N>
+where
+    I: MessageId,
+    P: Payload<MessageId = I, NodeId = N>,
+    N: NodeId,
+{
+    Gossip(GossipMsg<I, P, N>),
+    IHave(IHaveMsg<I, N>),
+    Graft(GraftMsg<I, N>),
+    Prune(PruneMsg<I, N>),
```
gorbak25, line 259: Comment what those messages do so our AI-agents doesn't need to read the original paper or infer their meaning from the codebase - will help in the future when digging in this code.

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+
+/// Digest-only batch sent to lazy peers on each tick.
+#[derive(Debug, Clone, Readable, Writable)]
+pub struct IHaveMsg<I, N>
+where
+    I: MessageId,
+    N: NodeId,
+{
+    pub sender: N,
+    pub digests: Vec<IHaveDigest<I>>,
```
gorbak25, line 283: How many digests are kept around here?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+    N: NodeId,
+    S: SeenStore<I>,
+> {
+    local_id: N,
+    config: Config,
+
+    eager_peers: HashSet<N>,
+    lazy_peers: IndexSet<N>,
+    known_peers: HashSet<N>,
+    ring_locked: HashSet<N>,
```
gorbak25, line 448: Peers who are always eager?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+    pub fn local_id(&self) -> &N {
+        &self.local_id
+    }
+
+    // --- Protocol methods ---
+
+    /// Handle an incoming GOSSIP message carrying a full payload.
+    ///
+    /// Unlike the paper, we do NOT promote the sender to eager on receipt.
+    /// In a multi-sender network, the peer that forwarded sender A's
```
gorbak25, line 587: How expensive would it be to maintain an separate broadcast tree per region 🤔 Would the per region trees look similar to the one this implementation creates.

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+            } else {
+                rt.notify(Notification::MessageMissing(ids.len()));
+                for id in ids {
+                    self.missing.remove(&id);
+                }
+            }
+        }
+    }
+
+    /// Periodic sends for all pending IHave digests to lazy peers.
```
gorbak25, line 957: When does a digest stop being pending? Do we wait for all lazy peers to ack?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+    }
+
+    fn commit_topology(&mut self, peer: N, info: RttInfo) {
+        self.peer_topology.insert(peer, info);
+        self.known_peers.insert(peer);
+        self.pending_topology.remove(&peer);
+    }
+
+    pub fn peer_down(&mut self, peer: &N, _rt: &mut impl Runtime<I, P, N>) {
+        let was_eager = self.eager_peers.remove(peer);
```
gorbak25, line 1107: Can we instantly promote an lazy pear to an eager one here when that happens?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+            self.eager_peers.insert(p);
+        }
+
+        let num_eager = self.num_eager();
+
+        // pools exclude ring_locked peers so they aren't reselected.
+        let (near_pool, mid_pool, far_pool) = self.bucket_pools();
+
+        // we mostly want eager peers to be low rtt, but throw in a mix of mid and far peers;
+        // percentage is roughly 50% near, 30% mid, 20% far. targets are computed over the
```
gorbak25: Can we make that ratio configurable and exposed in the config?

crates/plum-foca/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
+use indexmap::{IndexMap, IndexSet};
```
gorbak25, line 1: Amazing work! The main thing which is missing is to send PRUNE's to eager peers during graceful node shutdown. Imagine we run `systemctl stop corrosion2` on an node which has a lot of links. Without graceful handling we would need to first wait for foca to decect the peer is down, then plumtree will remove that peer and only after a few missed rounds the tree will get rebalanced. What is the round interval?

crates/corro-types/src/config.rs
```diff
@@ -211,6 +211,81 @@ pub enum AuthzConfig {
...
+pub fn default_plumtree_prune_threshold() -> u32 {
+    5
+}
+
+pub fn default_plumtree_optimization_threshold() -> u32 {
+    7
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PlumtreeConfig {
```
gorbak25, line 238: round interval?

crates/corro-types/src/broadcast.rs
```diff
@@ -89,6 +117,51 @@ pub enum AuthzV1 {
...
+    /// Application wants to originate a new broadcast.
+    Broadcast(ChangeV1),
+    /// Diagnostics: request a snapshot of the current plumtree state.
+    QueryStats(oneshot::Sender<PlumtreeStats>),
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct PlumtreeStats {
+    /// Peers we push GOSSIP to directly (the spanning-tree branches).
+    pub eager_peers: Vec<ActorId>,
```
gorbak25, line 172: OFFTOP: I'm tempted to vibe code a visualization of this :D Imagine we would query those edges every few seconds and we could lookup how the tree looks like

crates/corro-agent/src/plumtree/mod.rs
```diff
@@ -0,0 +1,1154 @@
...
+
+        if drop_oldest_plumtree_send(&mut p0_queue, &mut p1_queue, max_queue_len).is_some() {
+            log_at_pow_10(
+                "dropped old plumtree message from send queue",
+                &mut drop_log_count,
+            );
+            counter!("corro.plumtree.send.dropped").increment(1);
+        }
+    }
+
```
gorbak25, line 709: prune eager links to other peers or notify them that we're down?

gorbak25 (dismissed):
Amazing work! Mostly nits. The main thing i'm worrying about is that there are no tests for rebalancing the tree when a peer gets down/shuts down ungracefully. I don't see other peers getting notified that we're down on graceful shutdown.

doc/config/gossip.md
somtochiama: True. I'll also add a warning in logs when we receive the messages that don't correspond to the set broadcast algorithm

doc/config/gossip.md
somtochiama: Yes, Batching gossips didn't help smaller regional clusters. But I think it would be useful on global clusters that is pushing a lot of changes.

crates/plum-foca/tests/sim.rs
somtochiama: Yes, `Relative Message Redundancy (RMR)`  is from the plumtree paper, and it measures extra gossip received by nodes.

Most of the sim tests is AI generated. I'll go through and ensure the comments makes sense to someone looking thorugh corrosion.

crates/plum-foca/src/lib.rs
somtochiama: We store full payloads

crates/plum-foca/src/lib.rs
somtochiama: We currently don't config rounds, Rounds is mostly determined by fanout + optimization thresholds. A wider fanout means shorter rounds, a low optimization threshold means we'd try to graft a shorter path at a certain round.

crates/plum-foca/src/lib.rs
somtochiama: Yes, we recaclulate fanout as we receive peer_events so it grows with the cluster.

crates/plum-foca/src/lib.rs
somtochiama: Yes, this makes sure that a node always has two other peers that it gossips messages to regardless of the plumtree algorithm

crates/plum-foca/src/lib.rs
somtochiama: > When does a digest stop being pending? 

as soon as we send it

> Do we wait for all lazy peers to ack?

No acks, a lazy peer would send a GRAFT if it receives IHave for a message that it is missing. We do nothing if an ihave message is dropped, a node would either later receive it through gossip or during syncs

crates/corro-types/src/broadcast.rs
somtochiama: Ha, that's a very good idea. We could easily see the tree and how it is changing.

crates/plum-foca/src/lib.rs
somtochiama: fair, Mostly left membership notifications to foca.

Because tree is not symmetric - we could have eager peers that don't have us an eager peer. So prune only saves them from sending us a message. We could maintain a separate set with this info though, although foca is quick enough that I don't think that it is worth  it.

> plumtree will remove that peer and only after a few missed rounds the tree will get rebalanced.

Any missed messages would cause grafts before foca even sends the notifications.

crates/plum-foca/src/lib.rs
somtochiama: That sounds good

crates/plum-foca/src/lib.rs
somtochiama: This could make the tree less stable e.g we could select a recently demoted peer to be eager, and then prune them again etc.

We rely only on explicit grafts via missing messages + optimizations for eager peer selection.

crates/plum-foca/tests/sim.rs
somtochiama: duplicate_msgs also works. I'll change it

crates/plum-foca/tests/sim.rs
somtochiama: Yess. We can extend the test suite. I'll add it in

crates/plum-foca/src/lib.rs
somtochiama: A round is incremented every time a message is gossiped by a node to eager peers. We'd have to rebroadcast a message u32::max times before the counter overflows

gorbak25: Assuming we gossip 100 messages per second
```
>>> 2**32 / 100 / 3600 / 24
497.1026962962963
```
That comes out to ~1.5 years of continuous operation. Is that a big enough margin of error? What happens when that counter overflows?

somtochiama: The  round is per message and maximum round is roughly correlated to the depth of the tree. A cluster with a million nodes and a fanout of 3, would have a depth of about 14, accounting for duplicate nodes that would be selected during fanout, we could land on a max depth of 10*14 = 140

somtochiama: >  What happens when that counter overflows?

we could log an error, because that's mostly likely a  bug

crates/corro-types/src/config.rs
somtochiama: Do you mean number of rounds when we keep broadcasting a message?

crates/plum-foca/src/lib.rs
somtochiama: We don't have a specific number, it depends on the number of messages the sender node received within the ihave interval

crates/plum-foca/src/lib.rs
somtochiama: Updated code so we are now sending prunes to eager peers on shutdown

crates/plum-foca/src/lib.rs
somtochiama: It won't be expensive, and would look similar to what we have. We don't have the notion of a region in corrosion so a node would determine its regional peers based on RTT, then choose its eager and lazy peers from the set of low-RTT nodes..

The main thing is the sending changes to other regions, so they continue to be gossiped globally. Static gateway nodes aren't an option. One option is to gossip to randomly selected remote nodes? Feels like we might hit the same problems we had with normal gossip, but it could work

Another possibility is to retrofit gateway-like behavior into Plumtree itself. Each node could maintain a randomly selected peer from other inferred remote regions that are never pruned since they serve as cross-regional links. These gateway peers would only be reselected on peer_down events.

somtochiama:
Close #500

PR #495, opened 2026-07-10
Update uhlc, remove heavy features

Jake-Shadle:
The 0.7 version of uhlc was pulling in proc-macro-error2 which when compiled with rustc 1.97 issued a warning about containing code that will fail to compile in future versions, but then looking at defmt itself it made no sense to me why this feature was enabled in the first place as it's intended for `no_std` environments, removing the feature entirely removes several crates from the lockfile and doesn't cause any compilation or test failures. The only actual change introduced by the version bump is that `update_with_timestamp` returns a non-String error variant, but which implements Display.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1462,47 +1462,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "defmt"
-version = "0.3.100"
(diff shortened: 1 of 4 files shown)


somtochiama (approved):

somtochiama:
Thank you @Jake-Shadle !

PR #496, opened 2026-07-15
Add metrics for decompression

somtochiama:


crates/corro-types/src/broadcast.rs
```diff
@@ -110,7 +110,13 @@ pub fn compress_change(change: &ChangeV1, level: i32) -> Result<Vec<u8>, Compres
 }
 
 pub fn decompress_change(data: &[u8]) -> Result<ChangeV1, CompressError> {
+    let start = Instant::now();
     let decompressed = zstd::stream::decode_all(data)?;
+    histogram!("corro.decompression.time.seconds").record(start.elapsed());
+    if decompressed.len() > data.len() {
```
gorbak25, line 116: Is compression skipped when the ratio is > 1 ?

gorbak25 (approved):
lgtm

crates/corro-types/src/broadcast.rs
somtochiama: yes it is. but in that case we would send `ChangeV1` instead of `CompressedChange` so we still use decompressed bytes here?

PR #497, opened 2026-07-15
Allow zstd compression with custom dictionary

somtochiama:
This pull request;
-  adds the option of specifying a custom dictionary for  compressing / decompressing changes. This could improve compression ratio and speed up compression times. 
- adds command to write recently inserted changes from database to disk in speedy encoded format. The directory can then be passed to `zstd --train` to generate the custom dictionary.

### DIFF
--- a/crates/corro-types/src/config.rs
+++ b/crates/corro-types/src/config.rs
@@ -232,16 +232,46 @@ pub struct GossipConfig {
     pub disable_gso: bool,
     #[serde(default)]
     pub member_id: Option<MemberId>,
+    #[serde(default)]
+    pub compression: Option<CompressionConfig>,
+}
+
+impl GossipConfig {
+    pub fn compression_config(&self) -> CompressionConfig {
+        self.compression.clone().unwrap_or_default()
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CompressionConfig {
     /// Whether to zstd-compress broadcast and sync changeset payloads.
     /// Safe to flip cluster-wide at any time: uncompressed and compressed
-    /// nodes interoperate (broadcast drops what it can't decode and heals
-    /// via sync; sync negotiates per-peer via a capability flag).
+    /// nodes interoperate
     #[serde(default)]
-    pub compression: bool,
+    pub enabled: bool,
     /// zstd compression level for wire payloads (1–22). Ignored when
-    /// `compression` is false.
+    /// `enabled` is false.
     #[serde(default = "default_compression_level")]
-    pub compression_level: i32,
+    pub level: i32,
+    /// Path to a trained zstd dictionary (e.g. produced by `corrosion db
+    /// train-dict`), used to compress outgoing broadcasts. Its containing
+    /// directory is also scanned at startup for other trained dictionaries
+    /// (e.g. previous versions of it), which are loaded and indexed by
+    /// their embedded zstd dictionary id purely for decoding -- so
+    /// broadcasts from peers that haven't rotated to the current dictionary
+    /// yet can still be decoded.
+    #[serde(default)]
+    pub dict_path: Option<Utf8PathBuf>,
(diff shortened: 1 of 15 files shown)


crates/corrosion/Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 shellwords = { version = "1" }
 spawn = { path = "../spawn" }
+speedy = { workspace = true }
```
gorbak25, line 43: Have u explored https://github.com/rkyv/rkyv for this?

crates/corrosion/src/main.rs
```diff
@@ -850,6 +917,17 @@ enum TlsClientCommand {
 enum DbCommand {
     /// Acquires the lock on the DB
     Lock { cmd: String },
+    /// Write recent changes to disk as speedy-encoded `ChangeV1` samples,
+    /// one file per change chunk, for training a zstd compression
```
gorbak25: What's a chunk in that context?

crates/corrosion/src/main.rs
```diff
@@ -562,6 +565,70 @@ async fn process_cli(cli: Cli) -> eyre::Result<()> {
...
+                OpenFlags::SQLITE_OPEN_READ_ONLY
+                    | OpenFlags::SQLITE_OPEN_NO_MUTEX
+                    | OpenFlags::SQLITE_OPEN_URI,
+            )?)?;
+
+            std::fs::create_dir_all(out)?;
+            let tx = conn.transaction()?;
+
+            let mut version_stmt = tx.prepare(
+                "SELECT DISTINCT site_id, db_version
```
gorbak25, line 581: I believe this will result in an full scan of the clock tables as there's no index on ts. I think that's OK here as it's an admin command on an RO connection.

crates/corrosion/src/main.rs
somtochiama: Chunk in this context is `ChunkedChanges`. It is follows how corrosion sends out broadcast. I'll update the comment to just say this so it isn't confusing.

crates/corrosion/src/main.rs
somtochiama: Yes, this command isn't supposed to be run frequently. Just when the shape of data changes (clock tables changed, or we are writing to some new tables etc). We can also hook into current broadcasts that node is sending out but this is much simpler.

crates/corro-types/src/config.rs
```diff
@@ -232,16 +232,46 @@ pub struct GossipConfig {
...
+    /// yet can still be decoded.
+    #[serde(default)]
+    pub dict_path: Option<Utf8PathBuf>,
+}
+
+impl Default for CompressionConfig {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            level: default_compression_level(),
```
gorbak25, line 271: Can nodes coexist with different compression levels? on does this need to be the same in the entire cluster?

gorbak25 (approved):
lgtm

crates/corrosion/Cargo.toml
somtochiama: oh nice. I haven't looked at it yet. The admin command needs to use the same tool as corrosion, we want the same bytes that we pass to zstd. We can consider changing corrosion's serialization framework outside of this pr.

crates/corro-types/src/config.rs
somtochiama: Yes, they can coexist with different compression levels. 

What needs to be the same is the dictionary. The compressed bytes carries a dictionary id, and  we load every dictionary (with its id) in the same directory as the specified compressed file so a node can still decompress a change with a different dictionary.

somtochiama:
I am breaking the config format (putting compression settings in its own map). I think this is fine, since the old change is just one commit away and we haven't done any releases.

PR #499, opened 2026-08-03
Add command to reload zstd dicts

somtochiama:
This pr adds a command to load new zstd dictionaries in the compression directory into corrosion. This allows us to introduce a new dictionary without a restart


gorbak25 (approved):
lgtm!

PR #507, opened 2026-08-08
Prevent busy apply queue from starving buffered retries

tomr-a5:
A busy `rx_apply` queue can keep the biased select ready and prevent the five-minute retry tick from running. Failed fully buffered versions then stay at the initial timeout instead of being retried with the existing backoff.

This gives a due retry tick priority after shutdown and configures the interval to skip missed ticks, so one retry runs without a burst ahead of queued versions. The regression test covers a due retry and a ready version, then proves the queued version is preserved and handled next.

This is deliberately narrow. It does not change per-version backoff, timeout limits, SQL candidate selection, schemas, or APIs.

Validated with:

- `cargo test -p corro-agent due_retry_preempts_ready_new_version -- --nocapture`
- `cargo test -p corro-agent`
- `cargo clippy -p corro-agent --tests -- -D warnings`

### DIFF
--- a/crates/corro-agent/src/agent/util.rs
+++ b/crates/corro-agent/src/agent/util.rs
@@ -479,7 +479,9 @@ pub async fn apply_fully_buffered_changes_loop(
     let throttle_min = Duration::from_secs(5 * 60);
     let throttle_max = Duration::from_secs(60 * 60);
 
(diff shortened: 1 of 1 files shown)


tomr-a5:
Closing this after the incident gave us a better picture. I dont think this is the right direction anymore

PR #508, opened 2026-08-08
Fix Postgres transaction error handling

### LINKED ISSUE #502: Bulletproof postgres transaction handling
As per https://github.com/superfly/corrosion/pull/176 if an statement within an ongoing postgres transaction errors we should reject all commands other than a ROLLBACK. Currently we only do `discard_until_sync` on errors so if a transaction fails and the client sends a SYNC message then it can send another failing statement.
Also as per the original PR we should error on trying to start an nested transaction.

Guflly:
Keep explicit Postgres transactions in the failed state after an error so later commands return 25P02 until ROLLBACK. Also reject a nested BEGIN with 25001.

Adds coverage for both cases.

Fixes #502

### DIFF
--- a/crates/corro-pg/src/lib.rs
+++ b/crates/corro-pg/src/lib.rs
@@ -357,6 +357,7 @@ enum TxState {
     Started {
         kind: OpenTxKind,
         permits: Option<(OwnedSemaphorePermit, BookieWriteGuard)>,
+        failed: bool,
     },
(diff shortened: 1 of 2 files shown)


gorbak25 (changes requested):
CI fails.

gorbak25:
Supersceded by #520

PR #509, opened 2026-08-09
Expose tables through information_schema

### LINKED ISSUE #372: Emulate information_schema in the pg_listener
Right now tools like table plus are unable to know which tables are available when connecting via the PG listener.
Image
What currently works is that we are able to list the available databases.
Image

fzlzjerry:
## Summary

- attach an `information_schema` catalog for PostgreSQL listener sessions
- populate `information_schema.tables` directly from `sqlite_master`, including Corrosion and CR-SQLite internal tables
- expose PostgreSQL 14-compatible `information_schema.columns` metadata and primary-key rows through `information_schema.table_constraints`
- add PostgreSQL-wire integration coverage for user/internal table discovery, column metadata, and single/composite primary keys

Fixes #372

## Validation

- `cargo clippy -p corro-pg --all-targets -- -D warnings`
- `cargo nextest run --profile ci -p corro-pg`
- `git diff --check`

### DIFF
--- a/crates/corro-pg/src/lib.rs
+++ b/crates/corro-pg/src/lib.rs
@@ -80,7 +80,15 @@ use crate::{
     sql_state::SqlState,
     utils::CountedTcpStream,
     vtab::{
-        information_schema_tables::InformationSchemaTablesTable,
+        information_schema_columns::{
+            load_information_schema_columns, InformationSchemaColumnsTable,
+        },
+        information_schema_table_constraints::{
+            load_information_schema_table_constraints, InformationSchemaTableConstraintsTable,
+        },
+        information_schema_tables::{
+            load_information_schema_table_names, InformationSchemaTablesTable,
+        },
         pg_class::PgClassTable,
         pg_database::{PgDatabase, PgDatabaseTable},
         pg_namespace::PgNamespaceTable,
@@ -876,15 +884,12 @@ pub async fn start(
                         )?;
 
                         let dbs = Arc::new(vec![PgDatabase::new("state".into())]);
-                        let table_names = Arc::new(
-                            agent
-                                .schema()
-                                .read()
-                                .tables
-                                .keys()
-                                .cloned()
-                                .collect::<Vec<_>>(),
+                        let table_names = load_information_schema_table_names(&conn)?;
+                        let columns = Arc::new(load_information_schema_columns(&conn, &table_names)?);
+                        let table_constraints = Arc::new(
+                            load_information_schema_table_constraints(columns.as_slice()),
                         );
+                        let table_names = Arc::new(table_names);
 
                         conn.create_module(
                             "pg_database",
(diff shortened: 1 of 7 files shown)


crates/corro-pg/src/lib.rs
```diff
@@ -869,9 +870,21 @@ pub async fn start(
...
                         });
 
-                        conn.execute_batch("ATTACH ':memory:' AS pg_catalog;")?;
+                        conn.execute_batch(
+                            "ATTACH ':memory:' AS pg_catalog;
+                             ATTACH ':memory:' AS information_schema;",
+                        )?;
 
                         let dbs = Arc::new(vec![PgDatabase::new("state".into())]);
+                        let table_names = Arc::new(
```
gorbak25: This will skip the interesting `*__crsql_pks`, `__crsql_changes` tables. Just query `SELECT name FROM sqlite_master WHERE type='table'`

gorbak25 (changes requested):
While `information_schema.tables` is a nice first step we need also `information_schema.columns` to get external tooling working with the PG api. Possibly also `information_schema.table_constraints` with a `PRIMARY KEY` constrain.

crates/corro-pg/src/lib.rs
fzlzjerry: Updated this to read the table list directly from `sqlite_master`. The catalog now includes Corrosion and CR-SQLite internal tables such as `kitchensink__crsql_clock` and `kitchensink__crsql_pks`, with integration coverage for both.

fzlzjerry:
Addressed the requested catalog coverage in `ec858e1`:

- table names now come directly from `sqlite_master`, so internal Corrosion and CR-SQLite tables are included
- `information_schema.columns` now exposes the 44 PostgreSQL 14 fields, including nullability, defaults, PostgreSQL type names, ordinals, and generated-column metadata
- `information_schema.table_constraints` now reports primary keys, including composite primary keys
- the branch is rebased onto the latest `main`, and the full `corro-pg` nextest suite passes (12/12), along with formatting and clippy

gorbak25 (changes requested):
Almost here, TablePlus is able to list tables now 🎊 

Image

Unfortunately we need to list the tables under the "main" schema, not the "public" schema.
```
-- This works
SELECT count(*) FROM main.__corro_seq_bookkeeping;
-- This fails
SELECT count(*) FROM public.__corro_seq_bookkeeping;
```

Image

fzlzjerry:
Updated in `ffc1fbd`. The catalog now reports Corrosion tables under the SQLite `main` schema consistently:

- `information_schema.tables`, `columns`, and `table_constraints` use `main`
- `pg_catalog.pg_namespace` exposes `main` instead of the non-queryable `public` schema
- the integration test verifies the namespace and all three information-schema views

I rebased the change onto the branch update you pushed. `cargo nextest run -p corro-pg` passes 12/12, along with `cargo clippy -p corro-pg --all-targets -- -D warnings` and `cargo fmt --all -- --check`.

gorbak25 (approved):
Basic table listing in tableplus works now
Image

Remaining rough edges:
- Tableplus is unable to list columns
- Reading of tables with REAL types fails
-  Grafana is unable to enumerate any tables

Fixing the rough edges is out of the scope of this PR as the proper fix would entail at least:
- pg_attribute populated from the same source as information_schema.columns (Tableplus does `SELECT ... FROM information_schema.columns JOIN pg_attribute pa`)
- Populating pg_class
- Possibly mocking some function like `quote_ident` (https://github.com/grafana/grafana/blob/33dd204466841c3f5312d9f8ded0153d53084096/public/app/plugins/datasource/grafana-postgresql-datasource/postgresMetaQuery.ts)

PR #517, opened 2026-08-13
Re-queue buffered changes when schema is reloaded

somtochiama:


somtochiama:
closed in favour of #519

PR #519, opened 2026-08-18
Retry buffered partials after schema changes

### LINKED ISSUE #504: Retry applying fully buffered partial changesets on schema changes
Currently we retry the application of errored partial fully buffered change-sets from time to time with exponential exponential backoff. When the corrosion schema changes we should retry every partial regardless of it's current wait time to next application attempt.

onurata26:
Fixes #504.

## Summary

- Retry all fully buffered partials when the schema changes.
- Bypass the current retry wait while preserving the exponential backoff count.
- Avoid retries when a schema reload does not actually change the schema.
- Keep regular buffered-version retries subject to their existing backoff.

## Tests

- `cargo test -p corro-utils throttle_map_exponential_backoff_and_counts`
- `cargo test -p corro-agent test_apply_buffered_version_in_chunks`
- `cargo test -p corro-agent test_execute_schema_from_paths`
- `cargo test -p corro-agent test_schema_change_retries_all_fully_buffered_partials`
- `cargo clippy -p corro-utils -p corro-types -p corro-agent --all-targets -- -D warnings`
- `cargo check --workspace`

### DIFF
--- a/crates/corro-agent/src/agent/util.rs
+++ b/crates/corro-agent/src/agent/util.rs
@@ -504,7 +504,7 @@ pub async fn apply_fully_buffered_changes_loop(
                     vec![(actor_id, version)]
                 }
                 Some(ApplyTrigger::SchemaChanged) => {
-                    match find_fully_buffered_partials(&agent).await {
+                    match find_fully_buffered_partials(&agent, None).await {
                         Ok(partials) if partials.is_empty() => continue,
                         Ok(partials) => {
                             for partial in &partials {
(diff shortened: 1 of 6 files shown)


crates/corro-agent/src/agent/util.rs
```diff
@@ -566,6 +591,27 @@ pub async fn apply_fully_buffered_changes_loop(
     info!("fully_buffered_changes_loop ended");
 }
 
+async fn find_fully_buffered_partials(
```
somtochiama, line 594: Thanks for the pull request!. This looks good.

One nit: `find_fully_buffered_partials` and `find_fully_buffered_partial` are almost identical. We can have just one function that takes an optional limit .

onurata26: Done — I consolidated both lookups into a single helper with an optional limit. The schema-change path remains unbounded, while the periodic scan uses a limit of 1.

I also reran the regression test, the full corro-agent test suite, fmt, clippy, and workspace check locally.

somtochiama (approved):

PR #520, opened 2026-08-18
Fix Postgres transaction recovery semantics

### LINKED ISSUE #502: Bulletproof postgres transaction handling

onurata26:
Fixes #502.

This keeps an explicit Postgres transaction failed after a statement error until the client rolls it back. Subsequent commands return `25P02`, nested `BEGIN` returns `25001`, and extended-query errors wait for `Sync` before `ReadyForQuery`. `ROLLBACK` works through both simple and extended flows.

Transaction finalization now retains the write permit and bookkeeping guard until SQLite `COMMIT` or `ROLLBACK` has actually completed. Session teardown follows the same rule, so a failed cleanup cannot release the canonical writer early.

Regression coverage includes protocol recovery, finalization failures, rollback without leaked rows, and unit coverage for the `ConnectionDropWriteContext` ownership rule that holds the write permit until its final owner drops. [Fork validation](https://github.com/onurata26/corrosion/actions/runs/32595200234) for this head passed Linux/macOS workspace tests, lint, and book. The branch still needs upstream CI after it is brought up to date with `main`.

Related: #508.

### DIFF
--- a/crates/corro-pg/src/lib.rs
+++ b/crates/corro-pg/src/lib.rs
@@ -1294,7 +1294,8 @@ fn strip_pg_ast_expr(outer_expr: &mut sqlparser::ast::Expr) {
 
 #[cfg(test)]
 mod tests {
-    use super::pg_stmt_to_sqlite_sql;
+    use super::{pg_stmt_to_sqlite_sql, ConnectionDropWriteContext, TxState};
+    use std::{collections::HashMap, sync::Arc};
 
     #[test]
     fn check_normal_pg_dialect_array_subscript() {
@@ -1349,6 +1350,7 @@ mod tests {
         // Table references should keep pg_catalog. prefix since SQLite supports schema.table
         assert!(converted.contains("pg_catalog.pg_class"));
     }
+
     #[test]
     fn connection_drop_context_holds_write_permit_until_owner_drops() {
         let semaphore = Arc::new(tokio::sync::Semaphore::new(1));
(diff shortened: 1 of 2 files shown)


onurata26:
Found one more transaction-finalization edge case: `TxState` and the write guards were released before SQLite had actually finished `COMMIT` or `ROLLBACK`, so a failed finalization could leave SQLite in a transaction while Postgres reported `Idle`.

[`3863e62`](https://github.com/onurata26/corrosion/commit/3863e621a5b4ee2f439a83355e05176efdd2801b) keeps the transaction context until finalization succeeds or SQLite confirms the connection has returned to autocommit. [`9acd438`](https://github.com/onurata26/corrosion/commit/9acd43851381cca1176b94d5869cd6dc89951658) keeps it through connection teardown if rollback during session drop fails. The regression covers commit failure, `25P02`, rollback recovery without leaked rows, and permit cleanup.

Linux/macOS tests, lint, and the book build are green: [run 32595200234](https://github.com/onurata26/corrosion/actions/runs/32595200234).

somtochiama (approved):
LGTM! 

tagging @gorbak25 for a look too.

somtochiama:
Closes #508

gorbak25 (approved):
lgtm

PR #521, opened 2026-08-18
Deliver initial subscription snapshots directly

### LINKED ISSUE #183: Subscription channel can run out of capacity and drop events
The broadcast channel used by subscriptions (https://github.com/superfly/corrosion/blob/93561d180e3ebbdd275c9b80b9663b6dcb46e2fa/crates/corro-agent/src/api/public/pubsub.rs#L621) can run out of capacity and skip the events in case the data is read really fast from the disk.

Somehow, in this case, `forward_sub_to_sender` simply stops forwarding any event - https://github.com/superfly/corrosion/blob/93561d180e3ebbdd275c9b80b9663b6dcb46e2fa/crates/corro-agent/src/api/public/pubsub.rs#L728

It looks like the `select!` never matches `RecvError::Lagged` for some reasons.

onurata26:
Fixes #183.

The first subscriber now receives the matcher's initial snapshot directly, so a slow client cannot lose it to broadcast lag. Existing subscribers establish a committed catch-up boundary before the response and ignore live events already covered by catch-up. Matcher changes are published only after their SQLite transaction commits.

Tests cover initial backpressure and runtime errors, delayed broadcasts during catch-up, future cursors, and `skip_rows`. On current head `041adea`, the handoff regression passed repeatedly, resubscription and matcher tests passed, and Rust 1.89 clippy and formatting passed locally. Upstream CI is awaiting approval.

### DIFF
--- a/crates/corro-agent/src/api/public/pubsub.rs
+++ b/crates/corro-agent/src/api/public/pubsub.rs
@@ -633,14 +633,9 @@ pub async fn upsert_sub(
             return Err(MatcherUpsertError::SubFromWithoutMatcher);
         }
 
-        let (sub_tx, sub_rx) = broadcast::channel(10240);
+        let (sub_tx, sub_rx) = broadcast::channel(SUB_BROADCAST_CHANNEL_CAPACITY);
 
-        tokio::spawn(forward_sub_to_sender(
-            handle.clone(),
-            sub_rx,
-            tx,
-            params.skip_rows,
-        ));
+        tokio::spawn(catch_up_sub(handle.clone(), params, sub_rx, tx));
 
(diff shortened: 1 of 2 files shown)


onurata26:
The first-subscriber handoff now sends the initial snapshot directly and forwards startup errors. Added backpressure and runtime-error regression tests.

onurata26:
Follow-up in `041adea`: existing subscribers now use a committed catch-up boundary and ignore delayed broadcasts already covered by catch-up. Targeted tests, Rust 1.89 clippy, and formatting pass; upstream CI is awaiting approval.

gorbak25 (commented):
Looks good at first glance.
Will find some time this week to test it out on prod to see if there are some issues.

PR #522, opened 2026-08-18
Fix bundled SQLite and cr-sqlite integration

onurata26:
Follow-up to #414.

This fixes the bundled SQLite/cr-sqlite integration used by `custom-sqlite-bundle`. `crsql_bundle` is now a normal Rust dependency, while the C archive is kept whole so its Rust/C symbols remain linked. The vendored Rust code uses a host mode instead of installing its own allocator or panic handler, SQLite-owned values cross the boundary safely, and bundled `core_init` replaces runtime extension extraction.

The change also removes the default auto-vacuum setting that disables `BEGIN CONCURRENT`, adds a two-connection writer regression, and fixes nullable timestamps, zero-length blobs, and the vacuum/WAL fixtures.

[Run 32596152374](https://github.com/onurata26/corrosion/actions/runs/32596152374) tested source head [78b8618](https://github.com/onurata26/corrosion/commit/78b8618df2318dc9f5a15ddca525e8c92792ffc9): Linux and macOS each passed 71 tests with one skip; the existing macOS pubsub test passed on its third attempt, and docs passed. Linux clippy still fails on the pre-existing #414 warning at `crates/corro-pg/src/lib.rs:1072`; this PR does not touch that code.

Before this stack can land on `main`, refresh #414, then rebase/apply #522 and rerun full CI.

### DIFF
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -1091,7 +1091,7 @@ mod tests {
 
         {
             let db_conn = Connection::open(db_path.clone())?;
-            db_conn.execute_batch("PRAGMA auto_vacuum = INCREMENTAL")?;
+            db_conn.execute_batch("PRAGMA auto_vacuum = INCREMENTAL; VACUUM;")?;
         }
 
         println!("temp db: {db_path:?}");
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -442,6 +442,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.68.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "726e4313eb6ec35d2730258ad4e15b547ee75d6afaa1361a922e78e59b7d8078"
+dependencies = [
+ "bitflags 2.9.4",
+ "cexpr",
+ "clang-sys",
+ "lazy_static",
+ "lazycell",
+ "log",
+ "peeking_take_while",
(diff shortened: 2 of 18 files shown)


onurata26:
Updated the two bundled SQLite fixtures in [`78b8618`](https://github.com/onurata26/corrosion/commit/78b8618df2318dc9f5a15ddca525e8c92792ffc9): the incremental-vacuum test now runs `VACUUM` after setting `auto_vacuum=INCREMENTAL`, and the restore test checkpoints the source WAL before copying it. Both regressions pass on Linux and macOS (71 passed, 1 skipped on each): [run 32596152374](https://github.com/onurata26/corrosion/actions/runs/32596152374).

The run still fails Linux clippy on the inherited #414 `unnecessary_unwrap`; this PR does not touch that code. The macOS lint job was cancelled after the matrix failure. This remains a follow-up on `custom-sqlite-bundle`.

crates/corro-agent/src/agent/handlers.rs
```diff
@@ -1091,7 +1091,7 @@ mod tests {
 
         {
             let db_conn = Connection::open(db_path.clone())?;
-            db_conn.execute_batch("PRAGMA auto_vacuum = INCREMENTAL")?;
```
gorbak25, line 1094: Why?

gorbak25 (commented):
@onurata26 this is very needed but i don't have the capacity now to dig into this branch of work before #532 lands. I remember the last time i worked on this required https://github.com/superfly/cr-sqlite/pull/16 to be done and merged.

Issue #503, opened 2026-08-07
Sqlite writers concurrency for API/PG users


gorbak25:
Stock sqlite is limited to a single concurrent writer. This is problematic as one bad API client can block all remote sync processing and periodic maintenance we need to do. This makes the database stale and fall back behind it's peers. Fortunately readers can always access the current data and continue their work even under the presence of slow writers.

Sqlite has some experimental branches with some basic concurrency support like:
- [BEGIN CONCURRENT](https://www.sqlite.org/src/doc/begin-concurrent/doc/begin_concurrent.md)
- [hctree](https://sqlite.org/hctree/doc/hctree/doc/hctree/index.html)

Those are very experimental and are fairly limited, for instance the `BEGIN CONCURRENT` branch has page level concurrency control which makes it very impractical to be used as your concurrent transactions need to touch completely disjoint sets of pages which rarely happens. hctree tho is too early to be considered.

While at first glance this might hinder introducing concurrency we currently use cr-sqlite under the hood which already knows how to resolve conflicts between multiple concurrent writers. If we could "clone" the current database and execute our local changes against that clone then we could take it's changes and sync it back to the main database via cr-sqlite.

By using the experimental `BEGIN CONCURRENT` branch we can cheaply start multiple concurrent transactions each operating on the same concerete view of the DB and instead of committing their changes getting the `crsqlite_changes` which they made, rolling back the concurrent transaction and then syncing them back to the main db exactly the same way like we treat remote incoming changes.

To get this working we first need to build corrosion against a custom sqlite build with the concurrency patches AND at the same time bundling crsqlite directly into corrosion. To get cr-sqlite to build directly against corrosion means also making it working on stable rust.
The work on this has already been started via:
- Forking [sqlite-rs-embedded](https://github.com/superfly/sqlite-rs-embedded/commits/main/) and making it work on stable rust
- Making cr-sqlite work on stable rust - https://github.com/superfly/cr-sqlite/pull/16
- Bundling cr-sqlite + custom sqlite directly with corrosion - https://github.com/superfly/corrosion/pull/414

gorbak25:
A big complication which #523 exposed(thank you @onurata26 ❤️): https://sqlite.org/forum/info/a4c3a623f9ca999c

[...]

PR #523, opened 2026-08-19
Implement concurrent API and PG writers

onurata26:
Stacked on #522 and addresses #503.

Replayable API and Postgres writes now run in `BEGIN CONCURRENT`, capture their local `crsql_changes`, roll back the speculative transaction, and replay those changes under a short canonical write before broadcasting. DDL and other unsupported HTTP statements stay canonical, and PG transactions can upgrade without losing earlier CRR changes.

Replay is deliberately conservative. Targets are rechecked as live CRRs in the same SQLite snapshot, while user triggers, foreign keys, or temporary-schema state force canonical execution. If the schema version changes before replay, the transaction is rejected instead of being applied against a different schema. Failed explicit PG transactions accept only `ROLLBACK`; `COMMIT` remains `25P02`, nested `BEGIN` returns `25001`, and writer-permit ownership is retained through commit, rollback, and disconnect cleanup.

Known availability follow-ups are bounding the in-memory pending-change buffer and making canonical permit/replay waits cooperatively cancellable.

[Run 32598353717](https://github.com/onurata26/corrosion/actions/runs/32598353717) tested source head [69e9568](https://github.com/onurata26/corrosion/commit/69e9568bea05310bc25807d3fbdd1ab8f4591924) on Linux and macOS. Follow-up [8c5a021](https://github.com/onurata26/corrosion/commit/8c5a02133c62fdbeaadc6c7f65c4536a0e50304c) passed the targeted transaction regression, Rust 1.89 workspace clippy, and formatting locally; upstream CI has not started for this head.

Land #520 first, then refresh #414 and land #522. Rebase this PR onto that result, preserve #520's transaction-finalization behavior, drop the duplicate fixture commit `91d5806`, and rerun full CI before merging.

### DIFF
--- a/crates/corro-agent/src/agent/setup.rs
+++ b/crates/corro-agent/src/agent/setup.rs
@@ -66,6 +66,26 @@ pub struct AgentOptions {
     pub tripwire: Tripwire,
 }
 
+fn prepare_db_for_concurrent_writes(conn: &Connection) -> rusqlite::Result<()> {
+    let auto_vacuum: u64 = conn.pragma_query_value(None, "auto_vacuum", |row| row.get(0))?;
+
+    if auto_vacuum != 0 {
+        info!(
+            auto_vacuum,
+            "migrating database to auto_vacuum=NONE for concurrent writes"
+        );
+
+        conn.execute_batch(
+            "
+            PRAGMA auto_vacuum = NONE;
+            VACUUM;
+            ",
+        )?;
+    }
+
+    Ok(())
+}
+
 /// Setup an agent runtime and state with a configuration
 pub async fn setup(conf: Config, tripwire: Tripwire) -> eyre::Result<(Agent, AgentOptions)> {
     debug!("setting up corrosion @ {}", conf.db.path);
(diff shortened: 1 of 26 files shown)


onurata26:
Follow-up: failed transactions now stay aborted until an explicit `ROLLBACK`; `COMMIT` returns `25P02` in both protocol paths. Regression coverage also checks nested `BEGIN` and recovery. The previous full run covered `69e9568`; current head `8c5a021` passed the targeted test, workspace clippy, and formatting locally. Full CI will be rerun after the required rebase.

crates/corro-agent/src/agent/setup.rs
```diff
@@ -66,6 +66,26 @@ pub struct AgentOptions {
...
+
+    if auto_vacuum != 0 {
+        info!(
+            auto_vacuum,
+            "migrating database to auto_vacuum=NONE for concurrent writes"
+        );
+
+        conn.execute_batch(
+            "
+            PRAGMA auto_vacuum = NONE;
```
gorbak25, line 80: Incremental auto_vacuum is needed in production due to the db size. If having concurrency is mutually exclusive with auto_vacum=incremental then we need to investigate why and find a way forward.

PR #525, opened 2026-08-20
Drop plumtree mesages by priority/round

somtochiama:
This pull request adds a couple of fixes to the `send_messages_loop` in plumtree. The messages queue is split into local vs remote (so we can prioritize local gossip that haven't reached any nodes) and we now drop oldest messages first.

Also updates rtt measurement to take the maximum of the min_rtt that we've seen so far, so rtt measurements are more stable reducing plumtree rebalancing


gorbak25 (approved):
lgtm! Can the locked peers change? (like we find a better locked peer)

somtochiama:
@gorbak25 yes a locked peer can change when membership is updated (a peer is removed / added) and the adjacent peers (when sorted by actor_id) changes.

Ideally we'd want to lock down a ring0 peer tho cause those are peers we'd always send changes too, but because we rely on measurements to detect rtt node the link would be asymmetric? but that might be fine?

PR #526, opened 2026-08-21
Model check plum-foca with Stateright

### LINKED ISSUE #120: Explore _proving_ Corrosion's eventual consistency
There are tools like [`stateright`](https://github.com/stateright/stateright) to "prove" our model. I don't quite know if they're too involved or if they're easy to use.

onurata26:
Fixes #120.

This adds Stateright model checks for the real `PlumtreeState` rather than a parallel protocol implementation. The bounded three-node model explores delivery, timer, and lazy-push interleavings. It checks peer-set and delivery safety when one arbitrary protocol packet is dropped, and verifies recovery from a targeted loss of the initial eager Gossip. Recovery only advances after the faulted receiver requests a Graft, avoiding an unrelated node's request being treated as proof.

The model intentionally uses fixed membership and one broadcast to keep the state space manageable. Churn and broader loss patterns remain covered by the existing simulation tests and can be expanded separately.

Final CI passed on Linux and macOS (147/147 on each, plus lint and book): [run 32594591705](https://github.com/onurata26/corrosion/actions/runs/32594591705).


onurata26:
I tightened the packet-loss model after finding two ways recovery could be reported without proving the intended repair. Lossless delivery, arbitrary single-loss safety, and the targeted initial eager-Gossip loss are now separate scenarios. Targeted recovery only advances on a Graft from the faulted receiver, with a regression for an unrelated sender.

Linux/macOS validation is green (147/147 on each; lint and book passed): [run 32594591705](https://github.com/onurata26/corrosion/actions/runs/32594591705). The unrelated macOS pubsub test passed on retry; upstream CI still needs maintainer approval.

gorbak25 (changes requested):
@onurata26 Can u add an CI job which runs the model checker? @somtochiama do u think this provides value over the existing plumtree simulator test we have?

PR #527, opened 2026-08-21
Supervise long-running agent tasks

onurata26:
Addresses #108.

This PR adds supervision to selected long-running agent tasks. Services that can recover restart with bounded backoff; failures that cannot safely replay consumed work mark the agent unhealthy through the existing node recovery path.

**Changes and fixes**

- Supervise sync, membership notifications, gossip sending, incoming-change processing, buffered-change application, buffered-metadata cleanup, database maintenance, the configured Gossip/Plumtree broadcast loop, and agent/query metrics.
- Handle unexpected completion, returned errors, and panics using actor-specific recovery policies.
- Bound restart attempts with exponential backoff from 250 ms to 30 s. Reset the backoff and restart count after a stable run of at least 60 s; escalate after exhausting the ten-restart allowance.
- Keep channel receivers and relevant retry/queue state owned by actors across run invocations.
- Escalate unexpected receiver closure and replay-sensitive failures instead of repeatedly restarting a closed inbox or silently losing already-consumed work.
- Own Plumtree’s sender task in a `JoinSet`, observe unexpected completion, panic, and cancellation, and propagate failures to the supervisor. Drain the sender during normal shutdown and abort outstanding sender work on drop.
- Mark the agent unhealthy if an active change-batch task panics or is unexpectedly cancelled after leaving the queue. Abort and await active batch work when its actor exits.
- Own buffered-metadata cleanup jobs, observe their failures, and shut down outstanding jobs when the actor exits.
- Retain database-maintenance initialization and scheduling state, with shutdown-aware startup and timer waits.
- Ensure normal Tripwire shutdown does not trigger actor restart or escalation.
- Clamp the calculated batch threshold to at least one, preventing small batch configurations from rounding the threshold down to zero.
- Let the `SplitPool` dispatcher exit cleanly when all request channels close.
- Add actor restart/escalation counters and corresponding failure/restart logs.

**Integration with current main**

- Preserve #549’s range validation in `ChangesActor`, before length calculation, deduplication, clock/bookie access, rebroadcasting, or queueing. Regression coverage verifies that invalid changes are dropped while subsequent valid changes are processed.
- Integrate #551’s Plumtree selector and timer-priority changes with sender supervision. The order is shutdown, sender completion, protocol/IHAVE/maintenance timers, membership updates, then input.
- Preserve #551’s IHAVE missed-tick behavior and graceful protocol shutdown on channel closure.
[...]

PR #531, opened 2026-08-26
Evict dropped changes from the seen cache under their own actor id

### LINKED ISSUE #530: Load shed evicts seen cache entries under the wrong actor id
When the processing queue in `handle_changes` is full, the oldest queued change
gets dropped and its seen cache entry is supposed to be evicted, so a later
rebroadcast of that change can get back into the queue. The eviction uses the
wrong key:

```rust
while let Some(dropped_change) = state.maybe_drop_old_change() {
    for v in dropped_change.versions() {
        if let Entry::Occupied(mut entry) = seen.entry((change.actor_id, v)) {
```

`change` is the change that just arrived and triggered the drop. `v` iterates
the versions of the change being dropped. When those two come from different
actors, the wrong entry gets removed: the dropped change keeps its seen entry,
and the arriving actor can lose seen state for a same numbered version.

The first part is the bad one. Load shedding relies on rebroadcasts to repair
what it drops, and the stale entry blocks exactly those rebroadcasts: when
another peer re-sends the dropped change, the node filters it as a duplicate
and it never re-enters the queue. The node stays behind on that version until
the seen cache happens to trim the stale entry or a sync runs.

This came in with #431. `test_loadshed_handle_changes` doesn't catch it because
it runs a single actor, so the wrong key and the right key are the same value.
#527 reworks the surrounding code but keeps the same lookup.

Unit level repro: two actors, queue capacity 4. Actor A fills the queue with
partial changes (seqs 0..=3 of one version), actor B's arrival displaces A's
seq 0, then A's seq 0 is delivered again. On main the re-delivery is filtered
as a duplicate. I'll open a PR with a one line fix and a test for this.

## Cluster repro

To check this matters outside a unit test I ran it on a real 4 node cluster on
loopback. A writes 100 transactions of 40 rows each (one version per
transaction, splitting into about 5 wire chunks). B receives A's stream but is
frozen with SIGSTOP during the load phase and thawed after, so its rebroadcasts
arrive late, standing in for the redundancy a real fleet provides. C is the
victim: `processing_queue_len = 170` plus a loop of heavy local API writes that
hold the write connection, so the apply task starves and the queue sheds. D is
a second writer, so the evictions its arrivals trigger use D's actor id against
A's queued chunks. Sync backoff is set to 300s on all nodes so anti-entropy
cannot repair C inside the measurement window and mask the broadcast path.

B sometimes sheds a few of A's transactions itself while frozen, so the honest
measure is C against what B actually had to deliver, not against A. Five runs
per build, same parameters:

| run | main | with the fix |
|---|---|---|
| r0 | C 3760, B 3920, behind by 160 | C 3840, B 3840, behind by 0 |
| r1 | C 3840, B 4000, behind by 160 | C 4000, B 4000, behind by 0 |
| r2 | C 3800, B 4000, behind by 200 | C 3840, B 3840, behind by 0 |
| r3 | C 3800, B 4000, behind by 200 | C 4000, B 4000, behind by 0 |
| r4 | C 3840, B 4000, behind by 160 | C 3840, B 3840, behind by 0 |

Before B thaws, both builds shed about the same (dropped counters 121 to 144),
as expected. The difference is entirely in the retry wave: with the fix C
converges exactly to B's content in all runs, on main it never catches up
inside the window.

To reproduce, build a binary from main and one with the fix, then run the
script below with each:

```
./run_demo.sh /path/to/corrosion-main main
./run_demo.sh /path/to/corrosion-fixed fixed
```

Needs bash, curl and free ports 29011 to 29034 on loopback. Prints one RESULT
line per run.

<details>
<summary>run_demo.sh</summary>

```bash
#!/usr/bin/env bash
# Demo for the load-shedding seen-cache eviction bug (handlers.rs:1129).
#
# Topology: 3 corrosion nodes on loopback.
#   A - writer: streams small INSERTs into `tests`
#   B - relay: receives A's broadcasts, rebroadcasts them to C (late copies)
#   C - victim: tiny processing queue + its own heavy local API writes in
#       pulses, so the single write connection starves the broadcast-apply
#       task and the change queue sheds during pulses.
#
# Sync is pushed out to 300s on all nodes so the anti-entropy repair path
# cannot mask the broadcast-path behavior inside the demo window.
#
# Usage: ./run_demo.sh <corrosion-binary> <tag>
# Output: state + logs in state-<tag>/, summary line on stdout.
set -u

BIN="$1"; TAG="$2"
DIR="$(cd "$(dirname "$0")" && pwd)"
STATE="$DIR/state-$TAG"
N_TX=100         # multi-row txs from A; each ~40 rows x 800B = ~5 wire chunks
ROWS_PER_TX=40   # rows per tx (one version, many chunks, ONE seen-cache key)
PULSE_OFF=0.05   # C hammers itself almost continuously during the load phase
HEAVY_ROWS=8000  # rows per heavy local tx on C
N_TX_D=60
                 # evictions of A's queued chunks at C)
C_QUEUE=170
                 # < peak queued chunk items so chunks get shed

rm -rf "$STATE"; mkdir -p "$STATE/schema"
cat > "$STATE/schema/tests.sql" <<'EOF'
CREATE TABLE IF NOT EXISTS tests (
    text TEXT NOT NULL DEFAULT ""
) WITHOUT ROWID;

CREATE TABLE IF NOT EXISTS tests2 (
    text TEXT NOT NULL DEFAULT ""
) WITHOUT ROWID;
EOF

mk_config() { # name gossip_port api_port prom_port extra_perf
    local name=$1 gp=$2 ap=$3 pp=$4 extra=$5
    mkdir -p "$STATE/$name"
    cat > "$STATE/$name/config.toml" <<EOF
[db]
path = "$STATE/$name/corrosion.db"
schema_paths = ["$STATE/schema"]

[gossip]
addr = "127.0.0.1:$gp"
bootstrap = ["127.0.0.1:29011"]
plaintext = true

[api]
addr = "127.0.0.1:$ap"

[admin]
path = "$STATE/$name/admin.sock"

[telemetry]
prometheus.addr = "127.0.0.1:$pp"

[perf]
min_sync_backoff = 300
max_sync_backoff = 300
$extra
EOF
}

mk_config a 29011 29021 29031 ""
mk_config b 29012 29022 29032 ""
mk_config c 29013 29023 29033 "processing_queue_len = $C_QUEUE"
mk_config d 29014 29024 29034 ""

start_node() {
    local name=$1
    "$BIN" --config "$STATE/$name/config.toml" agent >"$STATE/$name/log" 2>&1 &
    echo $! > "$STATE/$name/pid"
}

api() { # port sql-json
    curl -s -m 10 -XPOST "http://127.0.0.1:$1/v1/transactions" \
        -H 'content-type: application/json' -d "$2" >/dev/null
}

count() { # port table -> prints count
    curl -s -m 10 -XPOST "http://127.0.0.1:$1/v1/queries" \
        -H 'content-type: application/json' -d "\"SELECT count(*) FROM $2\"" \
        | grep -o '"row":\[1,\[[0-9]*' | grep -o '[0-9]*$'
}

prom() { # port pattern
    curl -s -m 5 "http://127.0.0.1:$1" | grep -E "$2" | grep -v '^#'
}

cleanup() {
    for n in a b c d; do
        [ -f "$STATE/$n/pid" ] && kill "$(cat "$STATE/$n/pid")" 2>/dev/null
    done
    wait 2>/dev/null
}
trap cleanup EXIT

start_node a; start_node b; start_node c; start_node d
# wait for all APIs to come up
for port in 29021 29022 29023 29024; do
    for _ in $(seq 1 60); do
        curl -s -m 2 -XPOST "http://127.0.0.1:$port/v1/queries" \
            -H 'content-type: application/json' -d '"SELECT 1"' | grep -q values && break
        sleep 0.5
    done
done
sleep 3   # let gossip mesh form

# --- load phase ---
# Freeze the relay B for the whole load phase: its copies of A's changes
# become LATE copies, delivered only after C has shed and drained - exactly
# the retry a real fleet provides via gossip redundancy.
kill -STOP "$(cat "$STATE/b/pid")"

# A: N_TX multi-row transactions into `tests`. Each tx = one version whose
# changeset splits into ~5 8KB wire chunks: many queue items at C, one
# seen-cache key. A shed chunk leaves the whole version buffered (all its
# rows missing) until the relay's copy arrives.
TXT=$(printf 'x%.0s' $(seq 1 800))
mkdir -p "$STATE/a/payloads"
for i in $(seq 0 $((N_TX - 1))); do
    {
        printf '['
        for r in $(seq 1 $ROWS_PER_TX); do
            id=$((i * ROWS_PER_TX + r))
            [ "$r" -gt 1 ] && printf ','
            printf '"INSERT OR REPLACE INTO tests (id, text) VALUES (%d, %s)"' "$id" "'$TXT'"
        done
        printf ']'
    } > "$STATE/a/payloads/$i.json"
done
(
    seq 0 $((N_TX - 1)) | xargs -P 8 -I{} curl -s -m 15 \
        -XPOST "http://127.0.0.1:29021/v1/transactions" \
        -H 'content-type: application/json' \
        -d @"$STATE/a/payloads/{}.json" \
        -o /dev/null
) &
A_PID=$!

# D: second actor, multi-row txs into tests2 (distinct id range from C's
# hammer), paced across the load window. Its chunks interleave with A's in
# C's queue, so the evictions IT triggers hit A's chunks with D's actor id -
# the wrong-key case a single-writer topology can never produce.
mkdir -p "$STATE/d/payloads"
for i in $(seq 0 $((N_TX_D - 1))); do
    {
        printf '['
        for r in $(seq 1 $ROWS_PER_TX); do
            id=$((500000 + i * ROWS_PER_TX + r))
            [ "$r" -gt 1 ] && printf ','
            printf '"INSERT OR REPLACE INTO tests2 (id, text) VALUES (%d, %s)"' "$id" "'$TXT'"
        done
        printf ']'
    } > "$STATE/d/payloads/$i.json"
done
(
    for i in $(seq 0 $((N_TX_D - 1))); do
        kill -0 $A_PID 2>/dev/null || break
        curl -s -m 15 -XPOST "http://127.0.0.1:29024/v1/transactions" \
            -H 'content-type: application/json' \
            -d @"$STATE/d/payloads/$i.json" -o /dev/null
        sleep 0.1
    done
) &
D_PID=$!

# C: pulses of heavy local writes into `tests2` (hold the write connection).
# Payload built once (file - the statements exceed ARG_MAX); every iteration
# re-updates the same rows, which costs the same on the write connection.
{
    printf '['
    for k in $(seq 1 $HEAVY_ROWS); do
        [ "$k" -gt 1 ] && printf ','
        printf '"INSERT OR REPLACE INTO tests2 (id, text) VALUES (%d, %s)"' "$k" "'c-load'"
    done
    printf ']'
} > "$STATE/c/heavy.json"
(
    while kill -0 $A_PID 2>/dev/null; do
        curl -s -m 30 -XPOST "http://127.0.0.1:29023/v1/transactions" \
            -H 'content-type: application/json' -d @"$STATE/c/heavy.json" -o /dev/null
        sleep $PULSE_OFF
    done
) &
C_PID=$!

wait $A_PID
sleep 1
kill $C_PID $D_PID 2>/dev/null; wait $C_PID $D_PID 2>/dev/null

# let C fully drain while B is still frozen
sleep 6
ROWS_C_PRETHAW=$(count 29023 tests)
DROPPED_PRETHAW=$(prom 29033 'changes.*dropped|dropped' | awk '{s+=$NF} END {print s+0}')

# thaw B: it ingests A's stream now and rebroadcasts it to C - the late copies
kill -CONT "$(cat "$STATE/b/pid")"

# let the relay wave land; sync is 300s away so it cannot repair C
sleep 25

# --- measure ---
ROWS_A=$(count 29021 tests)
ROWS_B=$(count 29022 tests)
ROWS_C=$(count 29023 tests)
DROPPED=$(prom 29033 'changes.*dropped|dropped' | awk '{s+=$NF} END {print s+0}')
DUP_CACHE=$(prom 29033 'duplicate' | grep -i cache | awk '{s+=$NF} END {print s+0}')

SHED_LOG=$(grep -c 'dropped old change' "$STATE/c/log")
echo "RESULT tag=$TAG rows_a=${ROWS_A:-?} rows_b=${ROWS_B:-?} rows_c_prethaw=${ROWS_C_PRETHAW:-?} rows_c_final=${ROWS_C:-?} missing_c=$(( ${ROWS_A:-0} - ${ROWS_C:-0} )) dropped_prethaw=$DROPPED_PRETHAW dropped_total=$DROPPED shed_log_lines=$SHED_LOG dup_from_cache=$DUP_CACHE"
```

</details>

ZmoleCristian:
Fixes #530.

When the queue is full and `handle_changes` drops the oldest change, the seen
cache eviction looks up `(change.actor_id, v)`, where `change` is the change
that just arrived, while `v` iterates the versions of the change being
dropped. With more than one actor in the queue the dropped change keeps its
seen entry, so rebroadcasts of it get filtered as duplicates and the shed data
cannot re-enter the queue. Full analysis and a 4 node cluster repro are in the
issue.

The fix is to evict under `dropped_change.actor_id`.

The new test runs two actors: A fills the queue with partial changes, B's
arrival displaces A's seq 0, then A's seq 0 is delivered again and must be
accepted instead of being filtered as a duplicate. It fails on main and passes
with this change:

```
cargo test -p corro-agent --lib test_loadshed
```

One note on the test setup: it pins the batch size to 2 because with a min
batch size of 1 the batch threshold truncates to 0 and `handle_task_completion`
hits its `unreachable!()` on an empty queue. #527 already takes care of that
with `.max(1)`.

Happy to rebase if #527 lands first, the fix is the same line either way.

### DIFF
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -924,7 +924,7 @@ impl HandleChangesState {
     fn maybe_drop_old_change(&mut self) -> Option<ChangeV1> {
         let mut dropped_count = 0;
 
-        let maybe_dropped_change = if self.queue.len() >= self.max_queue_len {
+        if self.queue.len() >= self.max_queue_len {
             if let Some((dropped_change, _, _)) = self.queue.pop_front() {
                 self.buf_cost -= dropped_change.processing_cost();
                 dropped_count += 1;
@@ -937,9 +937,7 @@ impl HandleChangesState {
             }
         } else {
             None
-        };
-
-        maybe_dropped_change
+        }
     }
 
     /// Handle new change arrival - check if we should spawn immediately
@@ -1438,8 +1436,7 @@ mod tests {
         let partial = |actor: ActorId, version: u64, seq: u64| {
             let crsql_row = Change {
                 table: TableName("tests".into()),
-                pk: pack_columns(&vec![(seq as i64 + 1 + version as i64 * 100).into()])
-                    .unwrap(),
+                pk: pack_columns(&vec![(seq as i64 + 1 + version as i64 * 100).into()]).unwrap(),
                 cid: ColumnName("text".into()),
                 val: "partial".into(),
                 col_version: 1,
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -1126,7 +1126,7 @@ pub async fn handle_changes(
         // If we need to drop an old change, drop it from the seen cache aswell
         while let Some(dropped_change) = state.maybe_drop_old_change() {
             for v in dropped_change.versions() {
-                if let Entry::Occupied(mut entry) = seen.entry((change.actor_id, v)) {
+                if let Entry::Occupied(mut entry) = seen.entry((dropped_change.actor_id, v)) {
                     if let Some(seqs) = dropped_change.seqs() {
                         entry.get_mut().remove(seqs.into());
                     } else {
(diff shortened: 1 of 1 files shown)


gorbak25 (approved):
lgtm! Amazing work!

gorbak25:
@ZmoleCristian can u fix the lint failure? After that it should be good to merge.

ZmoleCristian:
On my way to the office 💪

ZmoleCristian:
> @ZmoleCristian can u fix the lint failure? After that it should be good to merge.

Should be good now I guess.

gorbak25 (approved):

PR #534, opened 2026-08-27
Rerun aggregate subscription queries on change

### LINKED ISSUE #326: Templates are rendered improperly when aggregations are used
On corrosion commit https://github.com/superfly/corrosion/commit/85a0aeddcdf479725b19e7cbf0edfb29c5444c6e with the following sql schema
```sql
CREATE TABLE foo (
    id INTEGER NOT NULL PRIMARY KEY
);
```
Templates using aggregation functions like json_group_array or sum render improperly
For ex. with the following rhai template
```
<%= sql("SELECT sum(foo.id) FROM foo").to_json(#{pretty: true}) %>
```
The result in an empty database looks like this
```
[
  {
    "sum(foo.id)": null
  }
]
```
Then after inserting some values `INSERT INTO foo(id) VALUES (1),(2),(3),(4);`
the result is
```
[
  {
    "sum(foo.id)": null
  },
  {
    "sum(foo.id)": 10
  }
]
```
Then after `INSERT INTO foo(id) VALUES (5);`
the result is
```
[
  {
    "sum(foo.id)": null
  },
  {
    "sum(foo.id)": 10
  },
  {
    "sum(foo.id)": 5
  }
]
```
While I would expect the template to render as
```
[
  {
    "sum(foo.id)": 15
  }
]
```

The same issue is present when using json_group_array with the following template `<%= sql("SELECT json_group_array(json_object('id', foo.id)) FROM foo").to_json(#{pretty: true}) %>
`

I would suggest to either disallow aggregations in the query parser or rerun the entire query when new data arrives.

samvallad33:
Aggregate templates like `<%= sql("SELECT sum(foo.id) FROM foo").to_json(#{pretty: true}) %>` render one row per changed primary key instead of a single rolled-up result. After `INSERT INTO foo(id) VALUES (1),(2),(3),(4)` you get stale null rows plus `{sum: 10}`; another insert adds `{sum: 5}` rather than `{sum: 15}`.

Subscriptions currently re-run the SELECT filtered to changed PKs. That works for row-shaped queries but breaks global aggregates (`sum`, `json_group_array`, etc.) which must see the whole table.

Detect aggregates at subscription parse time and rerun the full query on any matched table change, replacing prior result rows.

`cargo test -p corro-types test_aggregate_sum_rerun`

Fixes #326

### DIFF
--- a/crates/corro-types/src/pubsub.rs
+++ b/crates/corro-types/src/pubsub.rs
@@ -606,6 +606,7 @@ pub struct Matcher {
     pub cached_statements: HashMap<String, MatcherStmt>,
     pub pks: IndexMap<String, Vec<String>>,
     pub parsed: ParsedSelect,
+    has_aggregates: bool,
     pub evt_tx: mpsc::Sender<QueryEvent>,
     pub col_names: Vec<ColumnName>,
     pub last_rowid: u64,
@@ -885,6 +886,8 @@ impl Matcher {
 
         let processing_stats = handle.inner.processing_stats.clone();
 
+        let has_aggregates = parsed.has_aggregates;
+
         let matcher = Self {
             id,
             hash: sql_hash,
@@ -892,6 +895,7 @@ impl Matcher {
             cached_statements: statements,
             pks,
             parsed,
+            has_aggregates,
             evt_tx,
             col_names,
             last_rowid: 0,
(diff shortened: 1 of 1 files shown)


gorbak25:
@samvallad33 Subscriptions are meant to be incrementally maintainable. For example when handing a `SELECT count(*) FROM foo` subscription rerunning it every time when smth changes on the `foo` table would be very costly. For example imagine a table with 1kk rows and 1k updates per second - maintaining a count(*) subscription would be too costly. Because of this I won't merge this cause it will be too easy to shoot yourself in the foot.

The proper way to handle aggreagates is to compile them to some incrementally maintainable state and reconcile it on every change. For the count example u could keep an accumulator and add/subtract from it every time a row gets added/deleted.

The current corrosion subscription system needs a massive revamp. I've tackled that before in #406 but even that PR is not good enough - we first will probably need a better change distribution mechanism so the maintenance is cheaper.

PR #538, opened 2026-08-29
Make subquery child tables reactive in subscriptions

### LINKED ISSUE #537: Subscriptions ignore changes from tables referenced only inside subqueries
When a subscription references a table only inside `EXISTS`, `NOT EXISTS`,
`IN (SELECT ...)`, or a scalar subquery, Corrosion computes the initial result
correctly, but later changes to that table do not invalidate the subscription.

The SQL result changes while the open subscription emits no event and keeps its
cached result.

## Minimal reproduction

Schema:

```sql
CREATE TABLE services (
    marker INTEGER NOT NULL DEFAULT 0
) WITHOUT ROWID;

CREATE TABLE health_checks (
    service_id INTEGER NOT NULL,
    status TEXT NOT NULL
) WITHOUT ROWID;

CREATE INDEX health_checks_service_id
    ON health_checks (service_id);
```

Seed one service:

```sql
INSERT INTO services (id, marker) VALUES (1, 0);
```

Open a subscription to:

```sql
SELECT s.id
FROM services AS s
WHERE s.id = 1
  AND s.marker >= 0
  AND NOT EXISTS (
    SELECT 1
    FROM health_checks AS h
    WHERE h.service_id = s.id
      AND h.status != 'passing'
);
```

The initial snapshot contains service `1`. With the subscription still open,
insert a failing health check:

```sql
INSERT INTO health_checks (id, service_id, status)
VALUES (1, 1, 'failing');
```

Expected: the subscription emits a delete for service `1`.

Actual: a direct query returns no rows, but the subscription emits nothing.

The API sequence is:

```sh
URL=http://127.0.0.1:8080 # replace with the node's API address

# Keep this request open and wait for the initial row and eoq.
curl --no-buffer -X POST "$URL/v1/subscriptions" \
  -H 'content-type: application/json' \
  --data-binary "\"SELECT s.id FROM services AS s WHERE s.id = 1 AND s.marker >= 0 AND NOT EXISTS (SELECT 1 FROM health_checks AS h WHERE h.service_id = s.id AND h.status != 'passing')\""

# Run after eoq in another terminal.
curl -X POST "$URL/v1/transactions" \
  -H 'content-type: application/json' \
  --data-binary "[\"INSERT INTO health_checks (id, service_id, status) VALUES (1, 1, 'failing')\"]"

# This returns no row, while the subscription above emits no delete.
curl -X POST "$URL/v1/queries" \
  -H 'content-type: application/json' \
  --data-binary "\"SELECT s.id FROM services AS s WHERE s.id = 1 AND s.marker >= 0 AND NOT EXISTS (SELECT 1 FROM health_checks AS h WHERE h.service_id = s.id AND h.status != 'passing')\""
```

## Cause

Subquery parsing stores nested selects in `ParsedSelect.children`. Before the
patch, the matcher does not use those children when it builds its reactive
table map:

1. `Matcher::new` builds statements and dependency metadata from the root
   `ParsedSelect.table_columns` only.
2. `MatcherHandle::filter_matchable_change` checks incoming changes against
   that root map.
3. `health_checks` is absent, so its change is rejected before it can become a
   matcher candidate.
4. No candidate reaches `handle_candidates`, the cached query table is not
   diffed, and no subscription event can be emitted.

This is not a SQL evaluation or replication failure. The direct query observes
the new row immediately.

## Behavioral verification

I ran each case five times on fresh databases and fresh subscription stores.
Every trial checked the initial snapshot, required a quiet pre-mutation event
window, verified the post-mutation direct SQL result, and then classified the
subscription stream.

| Case | `main` at `936de111` | #406 at `0cf347d3` | With patch |
| --- | --- | --- | --- |
| Correlated `NOT EXISTS`, child insert/update/delete | stale, 5/5 each | stale, 5/5 each | reactive, 5/5 each |
| Correlated `EXISTS`, child insert/update | stale, 5/5 each | stale, 5/5 each | reactive, 5/5 each |
| `IN (SELECT ...)`, child insert | stale, 5/5 | stale, 5/5 | reactive, 5/5 |
| Uncorrelated `EXISTS`, child insert | stale, 5/5 | stale, 5/5 | reactive, 5/5 |
| Scalar subquery output update | stale, 5/5 | stale, 5/5 | reactive, 5/5 |
| Nested `EXISTS`, parent-child column update | stale, 5/5 | stale, 5/5 | reactive, 5/5 |
| Direct inner join control | reactive, 5/5 | reactive, 5/5 | reactive, 5/5 |
| Direct parent-table update control | reactive, 5/5 | reactive, 5/5 | reactive, 5/5 |

A clean release build from `936de111` reproduced the child-table failures in
5/5 trials.

PR #406 fixes aliased correlated parent references used only from a subquery
scope. At the tested commit, it still does not register subquery child tables
as reactive change sources.

## Cached-state probe

After the child mutation on `main`:

1. The direct query returned the new, correct empty result.
2. Opening the exact SQL again reused the same query id and returned the stale
   row.
3. Opening a semantically equivalent query with a different output alias
   created a new query id and returned the correct empty snapshot.
4. Updating a referenced `services` column made the original subscription emit
   the delayed delete.

This separates stale subscription state from SQL evaluation, transport, and
snapshot parsing.

## Two-node verification

A separate runner opened the subscription on node B and performed the mutation
on node A. It waited until a direct query on B proved that gossip had applied
the child row before inspecting the existing subscription on B.

| Build | Remote `NOT EXISTS` child insert | Remote inner join control |
| --- | --- | --- |
| `main` at `936de111` | stale, 5/5 | reactive, 5/5 |
| With patch | reactive, 5/5 | reactive, 5/5 |

The defect is downstream of replication.

The patch also passes the repository's complete local Linux CI path: rustfmt,
clippy across all targets with warnings denied, 155 workspace tests under the
`ci` nextest profile, the release binary build, and the mdBook build.

## Fix shape

The patch recursively collects two distinct sets from `ParsedSelect.children`.
Tables introduced by a child select's own `FROM` and `JOIN` clauses trigger a
full recomputation. All schema-table references found in child expressions
contribute reactive columns, including unaliased correlated outer references.
This lets nested parent-child column changes pass the change filter without
turning root-table changes into full recomputations. Batches containing only
root-table changes keep the existing primary-key-targeted path.

A child primary key does not generally identify the affected outer rows. A
delete makes this especially clear because the child row and its correlation
values are no longer present in current state.

## Performance characteristics

The full-result path was measured with one indexed `NOT EXISTS` subscription.
These values are matcher processing time for one child-driven batch, not a
request count or request duration.

| Result rows | Median processing time |
| ---: | ---: |
| 1,000 | 4.123 ms |
| 10,000 | 39.911 ms |
| 50,000 | 200.857 ms |
| 100,000 | 412.760 ms |
| 250,000 | 989.129 ms |

The measurement used a release build on an overclocked AMD Ryzen 7 5700G with
eight physical cores, SMT disabled, boost enabled, and a kernel-reported
maximum frequency of 5.658 GHz. Concurrent subscriptions and peak temporary
memory were not measured.

Candidate processing starts when the existing 600 ms buffer deadline expires
or when the buffer reaches 1,000 distinct table and primary-key entries.
Repeated changes to the same entry are coalesced, and one received candidate
set can carry a batch past the threshold. Each child-driven batch can cause one
full-result diff. Sustained change throughput was not measured, so these
single-batch timings do not establish a change-rate limit.

## Scope

This issue is limited to changes from tables referenced inside subqueries.

- Aliased correlated outer references used only from a subquery scope are
  addressed separately by #406. Unaliased schema-table references are retained
  as reactive columns by this patch without becoming full-recompute triggers.

ZmoleCristian:
Fixes #537.

Subquery parsing records nested selects in `ParsedSelect.children`, but the
matcher previously built its reactive table map from the root
`ParsedSelect.table_columns` only. A change from a table referenced only inside
a subquery was filtered before it could become a matcher candidate.

The initial subscription result was correct, but later child-table changes
could leave it stale indefinitely.

Full analysis, the behavioral matrix, and the two-node reproduction are in the
issue.

## What changed

- Recursively collect full-recompute triggers and reactive columns from nested
  selects.
- Use tables introduced by child `FROM` and `JOIN` clauses as full-recompute
  triggers. Keep every schema-table reference found in child expressions as a
  reactive column, including unaliased correlated outer references.
- Use the merged dependency map for change filtering and metrics. Record the
  same set in the existing `columns` metadata table, which is not read by the
  current matcher path.
- Run one full-result diff per buffered batch when the batch contains a
  subquery-table change.
- Keep the existing primary-key-targeted path for batches containing only
  root-table changes.
- Add regressions for a correlated `NOT EXISTS` child insert and trigger versus
  column classification across two nested child scopes.

## Why the subquery path uses a full diff

A changed child primary key does not generally identify the outer rows affected
by a subquery. For a delete, current state no longer contains the child row or
its correlation values.

Without provenance tracking or query-specific correlation analysis, targeting
the reevaluation by child primary key would be incomplete. The patch therefore
diffs the current full result against the cached query table once per buffered
child-driven batch.

## Regression test

The test starts with service `1` visible through a correlated `NOT EXISTS`
subscription. It inserts a failing health check and requires this exact event:

```text
Change(Delete, RowId(1), [1], ChangeId(1))
```

Without the implementation, the test times out because the `health_checks`
change is filtered. With the implementation, it passes. A parser-level test
also verifies that local child tables become full-recompute triggers while
unaliased correlated outer references remain reactive columns. The behavioral
matrix covers the nested parent-child column update in five isolated trials.

## Verification

```text
cargo fmt -- --check --color always
git diff --check
cargo test --locked -p corro-types --lib --release
  43 passed; 0 failed
RUSTFLAGS='--cfg tokio_unstable' cargo clippy --locked --all-targets -- -D warnings
  passed
RUSTFLAGS='--cfg tokio_unstable' cargo nextest run --profile ci --workspace \
  --target x86_64-unknown-linux-gnu
  155 passed; 0 failed; 10 skipped by profile
cargo build --locked --release --bin corrosion
  passed
mdbook build
  passed
```

Five isolated trials per build covered child inserts, updates, and deletes
through `EXISTS`, `NOT EXISTS`, `IN (SELECT ...)`, uncorrelated subqueries,
scalar subqueries, and a nested parent-child column reference. Direct inner
joins and parent-table changes were positive controls.

The same matrix was run against `main`, #406 at `0cf347d3`, and this patch. All
child-driven cases were stale on `main` and #406, and reactive with this patch.
The controls remained reactive on every build.

A separate two-node test wrote on node A and subscribed on node B. It first
verified through a direct query that gossip had applied the child row on B.
The existing subscription was stale on `main` in 5/5 runs and reactive with the
patch in 5/5 runs. A remote inner join control remained reactive.

## Performance characteristics

The full-result path was measured with one indexed `NOT EXISTS` subscription.
The values below are matcher processing time for one child-driven batch.

| Result rows | Median processing time |
| ---: | ---: |
| 1,000 | 4.123 ms |
| 10,000 | 39.911 ms |
| 50,000 | 200.857 ms |
| 100,000 | 412.760 ms |
| 250,000 | 989.129 ms |

The measurement used a release build on an overclocked AMD Ryzen 7 5700G with
eight physical cores, SMT disabled, boost enabled, and a kernel-reported
maximum frequency of 5.658 GHz.

Candidate processing starts when the existing 600 ms buffer deadline expires
or when the buffer reaches 1,000 distinct table and primary-key entries.
Repeated changes to the same entry are coalesced, and one received candidate
set can carry a batch past the threshold. Each child-driven batch can cause one
full-result diff.

The existing targeted path remains in use for root-table-only batches.
Sustained change throughput, concurrent subscription load, and peak temporary
memory were not measured. The single-batch timings do not establish a
change-rate limit.

## Scope

- Aliased correlated outer references used only from a subquery scope remain
  the concern of #406. Unaliased schema-table references remain on the targeted
  path in this patch.

### DIFF
--- a/crates/corro-types/src/pubsub.rs
+++ b/crates/corro-types/src/pubsub.rs
@@ -269,6 +269,7 @@ struct InnerMatcherHandle {
     hash: String,
     pool: sqlite_pool::RusqlitePool,
     parsed: ParsedSelect,
+    reactive_table_columns: IndexMap<String, HashSet<String>>,
     col_names: Vec<ColumnName>,
     cancel: CancellationToken,
     changes_tx: mpsc::Sender<MatchCandidates>,
@@ -366,8 +367,7 @@ impl Handle for MatcherHandle {
         // don't consider changes that don't have both the table + col in the matcher query
         if !self
             .inner
-            .parsed
-            .table_columns
+            .reactive_table_columns
             .get(change.table.as_str())
             .map(|cols| change.column.is_crsql_sentinel() || cols.contains(change.column.as_str()))
             .unwrap_or_default()
@@ -606,6 +606,8 @@ pub struct Matcher {
     pub cached_statements: HashMap<String, MatcherStmt>,
     pub pks: IndexMap<String, Vec<String>>,
     pub parsed: ParsedSelect,
+    reactive_table_columns: IndexMap<String, HashSet<String>>,
+    subquery_trigger_tables: IndexSet<String>,
     pub evt_tx: mpsc::Sender<QueryEvent>,
     pub col_names: Vec<ColumnName>,
     pub last_rowid: u64,
@@ -693,6 +695,16 @@ impl Matcher {
 
         if parsed.table_columns.is_empty() {
             return Err(MatcherError::TableRequired);
+        }
+
+        let subquery_trigger_tables = parsed.subquery_trigger_tables(schema);
+        let subquery_table_columns = parsed.subquery_table_columns(schema);
+        let mut reactive_table_columns = parsed.table_columns.clone();
+        for (table, columns) in &subquery_table_columns {
+            reactive_table_columns
+                .entry(table.clone())
+                .or_default()
+                .extend(columns.iter().cloned());
         }
 
         let mut statements = HashMap::new();
@@ -850,7 +862,7 @@ impl Matcher {
 
         // metrics counters
         let mut counter_map = HashMap::new();
-        for table in parsed.table_columns.keys() {
+        for table in reactive_table_columns.keys() {
             counter_map.insert(table.clone(), HandleMetrics{
                 matched_count: counter!("corro.subs.changes.matched.count", "sql_hash" => sql_hash.clone(), "table" => table.to_string()),
             });
@@ -867,6 +879,7 @@ impl Matcher {
                     .create_pool()
                     .expect("could not build pool, this can't fail because we specified a runtime"),
                 parsed: parsed.clone(),
+                reactive_table_columns: reactive_table_columns.clone(),
                 col_names: col_names.clone(),
                 cancel: cancel.clone(),
                 last_change_rx,
@@ -892,6 +905,8 @@ impl Matcher {
             cached_statements: statements,
             pks,
             parsed,
+            reactive_table_columns,
+            subquery_trigger_tables,
             evt_tx,
             col_names,
             last_rowid: 0,
@@ -1071,7 +1086,7 @@ impl Matcher {
             }
             trace!("created query indexes");
 
-            for (table, columns) in matcher.parsed.table_columns.iter() {
+            for (table, columns) in matcher.reactive_table_columns.iter() {
                 tx.execute(
                     r#"INSERT INTO columns ("table", cid) VALUES (?, '-1')"#,
                     [table.as_str()],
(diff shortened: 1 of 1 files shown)


gorbak25:
@ZmoleCristian This PR wil be left hanging without a review or a while because I need to determine whether this overlaps with #406

ZmoleCristian:
> @ZmoleCristian This PR wil be left hanging without a review or a while because I need to determine whether this overlaps with #406

Thanks for the heads-up.  🫡
I tested the minimal case against #406 at 0cf347d3, and it remained stale in 5/5 isolated runs while direct SQL and the controls were correct. 
I agree the implementation overlap is substantial, so I may be missing an intended path. 
Happy to share the reproducer or port the regression and fix onto #406 if useful.

PR #540, opened 2026-08-30
Reconcile LEFT JOIN subscriptions on child changes

### LINKED ISSUE #539: LEFT JOIN subscriptions miss null-side transitions on child changes
A subscription over a `LEFT JOIN` can become stale when a change comes from
the nullable-side table. Corrosion reevaluates rows targeted by the changed
child primary key, but the cached null-extended row has no child primary key
and falls outside that diff.

## Minimal reproduction

~~~sql
CREATE TABLE services (
    id INTEGER NOT NULL PRIMARY KEY
);

CREATE TABLE health_checks (
    service_id INTEGER NOT NULL,
    status TEXT NOT NULL
);

INSERT INTO services (id) VALUES (1);
~~~

Open a subscription to:

~~~sql
SELECT s.id
FROM services AS s
LEFT JOIN health_checks AS h
  ON h.service_id = s.id
 AND h.status != 'passing'
WHERE h.id IS NULL;
~~~

The initial snapshot contains service `1`. Keep the subscription open and run:

~~~sql
INSERT INTO health_checks (id, service_id, status)
VALUES (1, 1, 'failing');
~~~

Expected: the subscription emits a delete for service `1`.

Actual: a direct query returns no rows, but the subscription emits nothing.

The reverse transition fails as well. Updating or deleting the last matching
child makes the direct query return the null-extended row, but the
subscription does not emit the required insert.

For a projected query such as:

~~~sql
SELECT s.id, h.id
FROM services AS s
LEFT JOIN health_checks AS h
  ON h.service_id = s.id
 AND h.status != 'passing';
~~~

Corrosion emits the matched-row insert or delete but misses the corresponding
delete or insert of `(s.id, NULL)`.

## Cause

The child change reaches the matcher. The failure occurs inside the
primary-key-targeted reconciliation path:

1. The current-state query is restricted to the changed child primary key.
2. The existing optimization rewrites the targeted `LEFT JOIN` to
   `INNER JOIN`.
3. The cached side is restricted by the hidden child primary-key column.
4. A cached null-extended row has a `NULL` child PK, so it cannot match the
   changed child's real PK.
5. Neither side of the targeted diff contains the null-extended row.

Removing only the join rewrite is not sufficient. A child-PK slice still
cannot select a cached row whose child PK is `NULL`.

## Behavioral verification

Each trial used a fresh process, database, subscription store, loopback
address, and dynamic ports. The runner checked the initial snapshot, required
a quiet pre-mutation stream, verified the direct SQL result, and compared the
complete expected change set.

| Case | `main` at `936de111` | With patch |
| --- | --- | --- |
| Anti join, insert/update to matching | stale, 5/5 each | delete, 5/5 each |
| Anti join, update/delete from matching | stale, 5/5 each | insert, 5/5 each |
| Projected LEFT JOIN, null to matched | null-row delete missing, 5/5 | complete diff, 5/5 |
| Projected LEFT JOIN, matched to null | null-row insert missing, 5/5 | complete diff, 5/5 |
| Two matching children in one batch | null-row delete missing, 5/5 | complete diff, 5/5 |
| Root and child changes in one batch | child half missing, 5/5 | complete diff, 5/5 |
| Child on a second LEFT JOIN | null-row delete missing, 5/5 | complete diff, 5/5 |
| Inner join and root-update controls | reactive, 5/5 each | reactive, 5/5 each |
| Nonmatching child insert control | no change, 5/5 | no change, 5/5 |

A cache probe showed that reopening the same SQL reused the stale result, while
a fresh query id returned the correct snapshot. A later root-table update
repaired the original subscription. This was repeated in 5/5 unpatched trials.

## Two-node verification

A second runner wrote on node A and subscribed on node B. It waited for a
direct query on B to prove replication before checking the stream.

The remote anti join was stale in 5/5 runs on `main` and reactive in 5/5 runs
with the patch. The projected `LEFT JOIN` missed its null-row delete in 5/5
unpatched runs and emitted the complete diff in 5/5 patched runs. The remote
inner join control remained reactive.

## Fix shape

The patch records direct nullable-side `LEFT JOIN` tables during matcher
construction. When a buffered batch contains a change from one of those
tables, it diffs the full current subscription result against the full cached
result once. Root-only and inner-join batches retain the existing targeted
path. The patch also stops rewriting the now-unused child-targeted statement
to `INNER JOIN`.

A changed child PK cannot identify the null-extended cached row. Hidden root
PKs already exist in the cache, but they do not solve the minimal insert case:
the cached row has a `NULL` child PK and the current anti-join result restricted
by the new child PK is empty. A selective implementation would need to derive
affected root keys from the join inputs before the final query filter, or
retain another child-to-root mapping.

The regression test requires this exact event after inserting the matching
child:

~~~text
Change(Delete, RowId(1), [1], ChangeId(1))
~~~

It times out on unpatched `main` and passes with the implementation.

## Verification and cost

The repository's Linux CI path passes locally: rustfmt, clippy with warnings
denied, 154 workspace tests under the `ci` nextest profile, the release binary
build, and the mdBook build. The `corro-types` release test suite reports 42
passed.

Matcher-processing time for one indexed anti-join subscription:

| Result rows | Median processing |
| ---: | ---: |
| 1,000 | 4.132 ms |
| 10,000 | 39.790 ms |
| 50,000 | 200.300 ms |
| 100,000 | 403.754 ms |
| 250,000 | 1,018.066 ms |

The measurements used a release build on a shared development host with an
overclocked AMD Ryzen 7 5700G, eight physical cores, SMT disabled, boost
enabled, and a kernel-reported maximum frequency of 5.658 GHz. They do not
measure sustained write throughput, concurrent subscriptions, or peak
temporary memory.

Candidate batches are processed after the existing 600 ms deadline or at 1,000
distinct table and primary-key entries. Sustained nullable-side churn can
therefore cause repeated full-result work and growing matcher latency. The
candidate channel holds 20,480 items. On a full-channel `try_send`,
`match_changes` spawns a task that awaits `send` instead of deliberately
dropping the candidate. Overload behavior was not measured.

## Scope

This is independent of #537 and #538. Those cover tables referenced only
inside subqueries. Here the child table is a direct join dependency and its
change reaches the matcher, but the targeted diff cannot represent the
null-side transition.

`RIGHT JOIN` and `FULL JOIN` are outside this report and patch.

ZmoleCristian:
Fixes #539.

Subscriptions already treat a direct `LEFT JOIN` child as reactive, but the
child-PK-targeted diff cannot include a cached null-extended row. A matching
child insert can therefore leave an anti-join subscription stale. Projected
`LEFT JOIN` queries emit the matched-row half of the transition but miss the
null-row half.

Full reproduction and the repeated behavioral matrix are in the issue.

## What changed

- Record tables found on the nullable side of direct `LEFT` and
  `LEFT OUTER JOIN` clauses while building matcher statements.
- When a buffered batch contains a change from one of those tables, reconcile
  the complete subscription result once.
- Use the same augmented query shape as the initial snapshot and compare it
  with the complete cached `query` table.
- Reuse the existing insert, update, and delete classification.
- Keep the existing primary-key-targeted path for root-only and inner-join
  batches.
- Stop rewriting the now-unused nullable-child statement from `LEFT JOIN` to
  `INNER JOIN`.
- Add a regression for an anti-join child insert.

## Why this path uses a full diff

The cached null-extended row has a `NULL` hidden child primary key. It cannot
match the real primary key of a newly inserted, updated, or deleted child.
Removing the current `LEFT JOIN` to `INNER JOIN` rewrite does not fix that
slice boundary.

Hidden root PKs already exist in the cache, but they do not identify the root
for the minimal child-insert case: the cached row has a `NULL` child PK and the
current anti-join result restricted by the new child PK is empty. A selective
implementation would need to derive root keys from the join inputs before the
final query filter, or retain another child-to-root mapping. The complete diff
is the conservative correctness path with the metadata currently available.

If a batch contains both root and nullable-side child changes, the full diff
subsumes the root-targeted work. It runs once for that buffered batch.

## Regression test

`test_left_anti_join_reacts_to_child_insert` starts with service `1` visible,
inserts one matching child, and requires:

~~~text
Change(Delete, RowId(1), [1], ChangeId(1))
~~~

The test alone on unpatched upstream `main` times out after five seconds. With
the implementation, it receives the exact delete and passes.

## Behavioral verification

Five isolated trials per build covered:

- anti-join insert, update in both directions, and delete;
- projected null-to-matched and matched-to-null transitions;
- two matching children in one batch;
- a mixed root and child batch;
- a child change on a second `LEFT JOIN`;
- inner-join, root-update, and nonmatching-child controls.

All failing cases reproduced in 5/5 runs on `main` at `936de111` and were
correct in 5/5 runs with the patch. Every initial snapshot, direct SQL result,
and quiet pre-mutation window matched its expectation. All controls retained
their expected behavior.

A two-node runner wrote on node A, proved replication through a direct query on
node B, and then checked the subscription on B. The remote anti join and
projected `LEFT JOIN` failed in 5/5 unpatched runs and were correct in 5/5
patched runs. The remote inner join control remained reactive.

## Verification

~~~text
cargo fmt -- --check --color always
git diff --check
RUSTFLAGS='--cfg tokio_unstable' cargo test --locked -p corro-types --lib --release
  42 passed; 0 failed
RUSTFLAGS='--cfg tokio_unstable' cargo clippy --locked --all-targets -- -D warnings
  passed
RUSTFLAGS='--cfg tokio_unstable' cargo nextest run --locked --profile ci --workspace --target x86_64-unknown-linux-gnu
  154 passed; 0 failed; 10 skipped by profile
RUSTFLAGS='--cfg tokio_unstable' cargo build --locked --release --bin corrosion
  passed
mdbook build
  passed
~~~

I also ran the workspace through plain `cargo test`. The `corro-pg` integration
tests `test_pg_readonly` and `test_pg_corrrosion_shutdown` conflict over a
process-global tracing subscriber. The same integration-test binary fails with
the same two errors on an upstream-main worktree. Each test passes alone, and
both pass under the nextest CI run above.

I repeated the workspace run with `--no-fail-fast` so every standard test
target ran. `corro-pg::tests` was the only failed target, with 8 passed and the
same 2 failures. The same target reports 8 passed and 2 failed from a clean
main worktree.

## Performance characteristics

The full-result path was measured with one indexed anti-join subscription.
These are matcher-processing times for one child-driven batch:

| Result rows | Median | Observed range |
| ---: | ---: | ---: |
| 1,000 | 4.132 ms | 3.943 to 4.542 ms |
| 10,000 | 39.790 ms | 39.070 to 40.444 ms |
| 50,000 | 200.300 ms | 199.026 to 200.711 ms |
| 100,000 | 403.754 ms | 402.197 to 415.535 ms |
| 250,000 | 1,018.066 ms | 1,007.310 to 1,026.454 ms |

The measurement used a release build on a shared development host with an
overclocked AMD Ryzen 7 5700G, eight physical cores, SMT disabled, boost
enabled, and a kernel-reported maximum frequency of 5.658 GHz.

Candidate processing already starts after a 600 ms deadline or when the buffer
reaches 1,000 distinct table and primary-key entries. A nullable-side child
batch can cause one full-result diff. The candidate channel holds 20,480
items. When it is full, `match_changes` spawns an awaited send instead of
deliberately dropping that candidate. Sustained overload can still grow
matcher lag and the number of waiting tasks.

Sustained write throughput, concurrent subscription load, and peak temporary
memory were not measured.

## Interaction with #538

This branch is based directly on `main` and is independent of #538. Both
changes need a full-result path in `handle_candidates`. If #538 lands first,
this branch can reuse that path during rebase instead of retaining duplicate
control flow.

`RIGHT JOIN` and `FULL JOIN` remain outside this patch.

### DIFF
--- a/crates/corro-types/src/pubsub.rs
+++ b/crates/corro-types/src/pubsub.rs
@@ -606,6 +606,7 @@ pub struct Matcher {
     pub cached_statements: HashMap<String, MatcherStmt>,
     pub pks: IndexMap<String, Vec<String>>,
     pub parsed: ParsedSelect,
+    left_join_tables: IndexSet<String>,
     pub evt_tx: mpsc::Sender<QueryEvent>,
     pub col_names: Vec<ColumnName>,
     pub last_rowid: u64,
@@ -696,6 +697,7 @@ impl Matcher {
         }
 
         let mut statements = HashMap::new();
+        let mut left_join_tables = IndexSet::new();
 
         let mut pks = IndexMap::default();
 
@@ -776,18 +778,17 @@ impl Matcher {
                         Some(FromClause {
                             joins: Some(joins), ..
                         }) if idx > 0 => {
-                            // Replace LEFT JOIN with INNER join if the target is the joined table
+                            // Changes from the nullable side cannot be reconciled by child PK.
                             if let Some(JoinedSelectTable {
                                 operator:
                                     JoinOperator::TypedJoin {
-                                        join_type:
-                                            join_type @ Some(JoinType::LeftOuter | JoinType::Left),
+                                        join_type: Some(JoinType::LeftOuter | JoinType::Left),
                                         ..
                                     },
                                 ..
-                            }) = joins.get_mut(idx - 1)
+                            }) = joins.get(idx - 1)
                             {
-                                *join_type = Some(JoinType::Inner);
+                                left_join_tables.insert(tbl_name.clone());
                             };
 
                             // Remove all custom INDEXED BY clauses for the table as the most efficient
@@ -892,6 +893,7 @@ impl Matcher {
             cached_statements: statements,
             pks,
             parsed,
+            left_join_tables,
             evt_tx,
             col_names,
             last_rowid: 0,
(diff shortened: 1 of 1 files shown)


gorbak25:
@ZmoleCristian This PR wil be left hanging without a review or a while because I need to determine whether this overlaps with #406

PR #542, opened 2026-08-30
Recheck reaper candidates before deleting CR-SQLite metadata

### LINKED ISSUE #541: Reaper can delete CR-SQLite metadata for a row revived while it waits for the writer
When `[reaper]` is configured, `reap_table` selects expired deletion sentinels using a read-only connection and then waits for a low-priority write connection. The wait can be substantial when priority or normal writes are queued.

If a selected primary key is reinserted during that window, the current implementation still deletes all clock rows and the PK mapping for that key. The application row remains present, but its current CR-SQLite bookkeeping is gone.

The two phases currently have different predicates:

```sql
-- Read phase
WHERE col_name = -1
  AND col_version % 2 = 0
  AND ts < ?

-- Later write phase
WHERE key IN (previously_selected_keys)
```

A deterministic reproduction using the real reaper path and bundled CR-SQLite extension observed two clock rows created by the reinsertion being deleted.

Expected behavior: the write transaction should recheck that each selected key still has an expired deletion sentinel, including any configured `match_filter`. It should delete a PK mapping only if no clock rows remain after clock cleanup.

The issue is gated on reaper configuration, but the race window is realistic because the reaper deliberately uses the low-priority writer queue.

ZmoleCristian:
## Summary

- Reapply the deletion sentinel, version parity, retention cutoff, and `match_filter` predicates inside the reaper's immediate write transaction.
- Delete candidate PK mappings only when no clock rows remain.
- Add a deterministic regression that revives a row after read-side selection while the reaper waits for its low-priority writer.

The reaper previously selected keys on a read-only snapshot and later deleted bookkeeping rows by stale key alone. A reinsertion during the writer-queue wait could therefore lose its live clocks and PK mapping.

Candidate discovery stays on the read pool. The transaction-time query is bounded to the selected key list, preserving the existing row limit while closing the race.

## Tests

- `cargo test -p corro-agent agent::reaper::tests -- --nocapture`: 3 passed.
- `cargo nextest run -p corro-agent`: 51 passed, 2 skipped.
- `cargo fmt --all -- --check --color always`: passed.
- `cargo clippy -p corro-agent --all-targets -- -D warnings`: passed.
- `cargo clippy --all-targets -- -D warnings`: passed.
- `mdbook build`: passed with existing incomplete-link warnings.
- Independent `cargo nextest run --profile ci --workspace`: 154 passed, 10 skipped.

Closes #541

### DIFF
--- a/crates/corro-agent/src/agent/reaper.rs
+++ b/crates/corro-agent/src/agent/reaper.rs
@@ -145,6 +145,21 @@ async fn reap_table(
     clock: &uhlc::HLC,
     config: &ReaperCfg,
 ) -> Result<(usize, usize), ReaperError> {
+    reap_table_with_hook(pool, table, retention, filter, clock, config, || {}).await
+}
+
+async fn reap_table_with_hook<F>(
+    pool: &SplitPool,
+    table: &str,
+    retention: Duration,
+    filter: Option<&str>,
+    clock: &uhlc::HLC,
+    config: &ReaperCfg,
+    candidates_selected: F,
+) -> Result<(usize, usize), ReaperError>
+where
+    F: FnOnce(),
+{
     let read_conn = pool.read().await.map_err(ReaperError::ReadPool)?;
 
     let limit = config.row_limit;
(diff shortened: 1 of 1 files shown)


gorbak25 (approved):
That's a very unlikely race condition but yes - for correctness let's fix it

PR #546, opened 2026-08-30
fix: create backup parent directories before `VACUUM INTO`

### LINKED ISSUE #545: corrosion `backup` fails when the destination parent is missing
Backup fails unless every parent directory already exists.

While trying the documented backup flow with a dated destination path:

```css
corrosion backup /tmp/corrosion-backups/2026-08-31/snapshot.db
```

The command runs `VACUUM INTO` before `create_dir_all()`, so it exits with `unable to open database` before it reaches the directory creation code.

If I were to think, I can see why a backup tool might require the directory to exist, for example to avoid writing to local disk when an expected backup mount is missing. But corrosion already calls `create_dir_all()`, and I checked the git blame, the [commit that added it](https://github.com/superfly/corrosion/commit/7b0b4fc9a05a27d08be4c592b0f6a935225d40ee) says `create backup parent directory automatically`. I would imagine that this would keep that existing behavior and run it before the operation that needs the directory. Please correct me if it's intentional.

I saw [#105](https://github.com/superfly/corrosion/issues/105) is vaguely related cause it discusses replacing the same `VACUUM INTO` step with SQLite's backup API. That may eventually replace this code path, but it doesn't address the current directory ordering failure. Either implementation still needs a valid destination parent.

wthrajat:
Attempts to fix #545 in case the whole thing isn't intentional (hence keeping the PR draft for now)

Also, the test creates an SQLite db with one row and backs it up to a nested path whose parent does not exist. It then opens the backup and checks that the row was copied.

```bash
cargo fmt --all -- --check
cargo test -p corrosion --locked
cargo clippy -p corrosion --bin corrosion --tests --locked --no-deps -- -D warnings
```

### tldr

- missing destination directories are created first
- `VACUUM INTO` produces the backup
- directory creation errors are returned instead of ignored
- added a regression test
- all corrosion tests pass


gorbak25 (approved):
Thank you!

PR #547, opened 2026-08-31
fix: bump tokio-postgres/postgres-protocol to patch RUSTSEC-2026-0178/0179/0180

minutechreview:
## What

`crates/corro-pg` runs a real, network-reachable Postgres wire-protocol server (`TcpListener` in `crates/corro-pg/src/lib.rs`). The locked `tokio-postgres 0.7.13` / `postgres-protocol 0.6.8` carry three DoS advisories:

- **RUSTSEC-2026-0178** (tokio-postgres): panic on a malformed `DataRow` with fewer fields than columns.
- **RUSTSEC-2026-0179** (postgres-protocol): unbounded SCRAM iteration count → CPU-exhaustion DoS.
- **RUSTSEC-2026-0180** (postgres-protocol): panic decoding a malformed `hstore` payload.

All three are fixed upstream (`tokio-postgres` 0.7.18, `postgres-protocol` 0.6.12), and both are already within the existing `Cargo.toml` semver range (`tokio-postgres = { version = "0.7.12" }`), so this is a `Cargo.lock`-only change via `cargo update -p tokio-postgres -p postgres-protocol` — no manifest edits, no API surface changes.

## Why now

Checked first: no `dependabot.yml` in this repo, and no existing issue or PR addresses this.

## Verification

- `cargo build -p corro-pg` — clean, no warnings or errors.
- `cargo test -p corro-pg` — 8 passed / 2 failed, both with and without this change (verified side-by-side in a separate worktree on unmodified `main`). The two failures (`test_pg_readonly`, `test_pg_corrrosion_shutdown`) are pre-existing test-isolation issues — a global `tracing_subscriber::set_global_default` call and a `std::sync::Once` that only the first test in the binary to reach them can succeed at — identical panic messages on both the base commit and this change. Not a regression from this bump.

## Notes

- Toolchain used: `rustc 1.89.0` (repo's pinned `rust-toolchain.toml`), via `rustup`.
- Ran `cargo update --dry-run` first to confirm the exact resolved versions before applying, and diffed `Cargo.lock` to confirm only the expected 24 packages moved (transitive updates pulled in alongside the two direct fixes — full list in the commit message / lockfile diff).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2663,7 +2663,7 @@ checksum = "e04d7f318608d35d4b61ddd75cbdaee86b023ebe2bd5a66ee0915f0bf93095a9"
 dependencies = [
  "hermit-abi 0.5.2",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4584,7 +4584,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5131,7 +5131,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "psm",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -456,6 +456,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.12.1"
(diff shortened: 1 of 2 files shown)


gorbak25 (dismissed):
@minutechreview go ahead and bump the versions in cargo.toml aswell

minutechreview:
Done — bumped the dev-dependency's declared `tokio-postgres` floor from 0.7.12 to 0.7.18 in `crates/corro-pg/Cargo.toml` (only direct declaration of it anywhere in the repo; `postgres-protocol` is purely transitive via `pgwire`, no manifest line to bump there). `cargo check --all-targets` is clean on the updated workspace.

PR #549, opened 2026-08-31
Reject inverted synchronization ranges before processing

### LINKED ISSUE #548: Inverted sync ranges can panic synchronization and change handling
Corrosion accepts inclusive ranges from synchronization messages without checking that the start is less than or equal to the end.

This affects three serialized inputs:

- `SyncStateV1.need` and `SyncStateV1.partial_need`
- `SyncNeedV1::Full` and `SyncNeedV1::Partial`
- the version or sequence ranges carried by `ChangeV1`

An inverted range survives Speedy decoding. It can then reach `RangeInclusiveSet` while Corrosion computes available sync work or updates the recently-seen change cache. `rangemap` asserts on that input, so the task handling the value exits instead of returning a validation error.

A minimal regression constructs a range from 10 through 5, round-trips it through the existing serialized type, and passes the decoded value into the same range operations used by the sync and change handlers. The baseline accepts the encoding and reaches the assertion. Valid ranges remain unaffected.

Expected behavior: serialized values containing inverted inclusive ranges should be rejected before any range-set operation. Values constructed inside the process should return a typed error, or be logged and dropped, at the processing boundary without terminating the task.

ZmoleCristian:
## Summary

- Validate sync-state, sync-request, and change ranges as they are decoded.
- Return typed errors when in-memory sync state reaches `compute_available_needs` with an invalid range.
- Validate every aggregated in-memory sync request before actor and range filtering.
- Recheck requests at `handle_need` for callers that bypass the normal batch path.
- Log and drop invalid queued changes before they reach metrics, deduplication, clock handling, or the seen cache.

The custom Speedy readers decode into private unvalidated representations and construct the existing public types only after validation. Their field order, enum variant order, and minimum-size hints remain unchanged, so valid serialized messages keep the same representation.

`ChangeV1` validation covers sequence ranges in full changesets and every version range in empty and empty-set changesets.

## Tests

- Added rejection tests for full and partial sync state.
- Added rejection tests for full and partial sync requests.
- Added sync computation controls for valid and invalid in-memory state.
- Added sync and broadcast tests for invalid change version and sequence ranges.
- Added valid round-trip and seen-cache controls.
- Added a batch preflight control that validates all direct requests before filtering.
- Compared valid state, request, and change encodings from the exact base and fixed source byte-for-byte.
- `cargo test --locked -p corro-types range_validation_tests`
- `cargo test --locked -p corro-agent range_validation_checks_all_direct_needs_before_processing`
- `cargo clippy --locked --all-targets -- -D warnings`
- `cargo nextest run --locked --profile ci --workspace --target x86_64-unknown-linux-gnu`
- `mdbook build`

Results: 14 of 14 focused tests passed. Valid baseline and fixed encodings produced the same SHA-256. The exact CI run passed 167 of 167 tests with 10 upstream skips; one existing pubsub test passed on its second configured attempt. Clippy, formatting, mdBook, linkcheck, and diff checks passed.

## Compatibility

Valid messages are unchanged on the wire. The behavioral change is limited to rejecting ranges whose start is greater than their end.

Closes #548

### DIFF
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -1055,6 +1055,11 @@ pub async fn handle_changes(
             }
         };
 
+        if let Err(error) = change.validate() {
+            warn!(actor_id = %change.actor_id, %error, "dropping change with an invalid range");
+            continue;
+        }
+
         let change_len = change.len();
         counter!("corro.agent.changes.recv").increment(std::cmp::max(change_len, 1) as u64); // count empties...
 
--- a/crates/corro-agent/src/api/peer/mod.rs
+++ b/crates/corro-agent/src/api/peer/mod.rs
@@ -15,8 +15,9 @@ use corro_types::broadcast::{
 use corro_types::change::{row_to_change, Change, ChunkedChanges};
 use corro_types::config::GossipConfig;
 use corro_types::sync::{
-    generate_sync, SyncMessage, SyncMessageEncodeError, SyncMessageV1, SyncNeedV1, SyncRejectionV1,
-    SyncRequestV1, SyncStateV1, SyncTraceContextV1,
+    generate_sync, SyncMessage, SyncMessageEncodeError, SyncMessageV1, SyncNeedV1,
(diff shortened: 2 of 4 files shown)


gorbak25 (approved):
Love it!

PR #551, opened 2026-08-31
Prevent Plumtree timers from starving behind inbound traffic

### LINKED ISSUE #550: Plumtree timers starve under continuous inbound traffic
`plumtree_loop` uses a biased `tokio::select!`, but both inbound channels are polled before the protocol and maintenance timers.

The current order allows a continuously ready `rx_plumtree` channel to prevent all lower branches from running. A saturated three-second probe handled more than one million inputs while handling zero GRAFT timers and zero IHAVE ticks.

The affected branches are responsible for:

- firing GRAFT repair after an IHAVE-announced message remains missing;
- flushing queued IHAVE digests every 150 ms;
- updating peer topology, evicting payload/seen caches, and recording gauges every 60 s.

This means sustained inbound traffic can suppress tree repair and lazy propagation. It can also leave the payload and seen caches above their configured targets until the input channel becomes temporarily empty, because eviction only runs from the maintenance branch.

A deterministic reproduction against the production event-loop selector makes the input receiver and a protocol timer ready simultaneously. The input branch wins with the current arm order. The same behavior applies to the lower IHAVE and maintenance branches.

Expected behavior: Plumtree's protocol and maintenance timers should be serviced even while inbound traffic remains continuously available.

ZmoleCristian:
## Summary

- Prioritize Plumtree protocol, IHAVE, and maintenance timers before inbound channels in the biased event-loop selector.
- Keep graceful shutdown at the highest priority.
- Extract the selector so deterministic tests exercise the exact arbitration used by `plumtree_loop`.

Previously, a continuously ready Plumtree input channel prevented every lower timer branch from being polled. That suppressed missing-message GRAFT repair, IHAVE flushing, topology maintenance, cache eviction, and related gauges until an input gap occurred.

The input and update handlers, timer durations, protocol state transitions, and channel capacities are unchanged.

The selector intentionally remains priority-based rather than round-robin: each iteration handles one selected timer or tick, then polls again. This prevents hot inbound channels from starving ready timers without claiming a latency bound for draining an existing timer backlog.

## Tests

- Added `plumtree_timers_preempt_ready_input`, covering protocol, IHAVE, and maintenance timers against a simultaneously ready input.
- `cargo test -p corro-agent plumtree_timers_preempt_ready_input -- --nocapture`
- `cargo fmt -- --check --color never`
- `RUSTFLAGS='--cfg tokio_unstable' cargo clippy --all-targets -- -D warnings`
- `RUSTFLAGS='--cfg tokio_unstable' cargo nextest run --profile ci --workspace --target x86_64-unknown-linux-gnu`

The exact CI nextest run and an independent final rerun each completed with 154 passed and 10 skipped by the repository configuration.

Fixes #550

### DIFF
--- a/crates/corro-agent/src/plumtree/mod.rs
+++ b/crates/corro-agent/src/plumtree/mod.rs
@@ -29,7 +29,7 @@ use speedy::Writable;
 use tokio::{
     sync::mpsc,
     task::JoinSet,
-    time::{interval, MissedTickBehavior},
+    time::{interval, Interval, MissedTickBehavior},
 };
 use tokio_util::codec::{Encoder, LengthDelimitedCodec};
 use tracing::{error, info, trace, warn};
@@ -380,6 +380,46 @@ pub async fn spawn_plumtree_loop(
     .await;
 }
 
+enum PlumtreeLoopBranch {
+    Shutdown,
+    Input(PlumtreeInput),
+    InputClosed,
+    Updates(PlumtreeUpdates),
+    UpdatesClosed,
+    HandleTimer(Timer<ChangeId, ActorId>),
+    IHaveTick,
+    MaintenanceTick,
+}
+
+async fn next_plumtree_loop_branch(
+    tripwire: &mut Tripwire,
+    rx_plumtree: &mut CorroReceiver<PlumtreeInput>,
+    rx_plumtree_updates: &mut CorroReceiver<PlumtreeUpdates>,
+    plumtree_timer_rx: &mut mpsc::Receiver<(Timer<ChangeId, ActorId>, Instant)>,
+    ihave_tick_interval: &mut Interval,
+    maintenance_interval: &mut Interval,
+) -> PlumtreeLoopBranch {
+    tokio::select! {
+        biased;
+        _ = tripwire => PlumtreeLoopBranch::Shutdown,
+        // These timers drive tree repair and cache maintenance. Keep them
+        // ahead of inbound channels, which can remain continuously ready.
+        Some((timer, _seq)) = plumtree_timer_rx.recv() => {
+            PlumtreeLoopBranch::HandleTimer(timer)
+        }
+        _ = ihave_tick_interval.tick() => PlumtreeLoopBranch::IHaveTick,
+        _ = maintenance_interval.tick() => PlumtreeLoopBranch::MaintenanceTick,
+        updates = rx_plumtree_updates.recv() => match updates {
+            Some(updates) => PlumtreeLoopBranch::Updates(updates),
+            None => PlumtreeLoopBranch::UpdatesClosed,
+        },
+        input = rx_plumtree.recv() => match input {
+            Some(input) => PlumtreeLoopBranch::Input(input),
+            None => PlumtreeLoopBranch::InputClosed,
+        },
+    }
+}
+
 pub async fn plumtree_loop(
     agent: Agent,
     transport: Transport,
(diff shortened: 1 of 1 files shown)


gorbak25 (approved):
Makes sense but let's @somtochiama weigh in before merging

crates/corro-agent/src/plumtree/mod.rs
```diff
@@ -380,6 +380,46 @@ pub async fn spawn_plumtree_loop(
...
+        biased;
+        _ = tripwire => PlumtreeLoopBranch::Shutdown,
+        // These timers drive tree repair and cache maintenance. Keep them
+        // ahead of inbound channels, which can remain continuously ready.
+        Some((timer, _seq)) = plumtree_timer_rx.recv() => {
+            PlumtreeLoopBranch::HandleTimer(timer)
+        }
+        _ = ihave_tick_interval.tick() => PlumtreeLoopBranch::IHaveTick,
+        _ = maintenance_interval.tick() => PlumtreeLoopBranch::MaintenanceTick,
+        updates = rx_plumtree_updates.recv() => match updates {
```
somtochiama, line 412: Thank you for the PR! 
Let's have `rx_plumtree` before `rx_plumtree_update`, the maintenance tick also updates plumtree with the membership state so it is okay the channel is starved a little.

crates/corro-agent/src/plumtree/mod.rs
```diff
@@ -380,6 +380,46 @@ pub async fn spawn_plumtree_loop(
     .await;
 }
 
+enum PlumtreeLoopBranch {
+    Shutdown,
+    Input(PlumtreeInput),
+    InputClosed,
+    Updates(PlumtreeUpdates),
+    UpdatesClosed,
```
somtochiama, line 388: Small nit: Updates closed / InputClosed / Shutdown roughly do the same thing, i.e log, run handle shutdown and  exit the loop (I think we should `state.handle_shutdown(&mut rt);` in all cases). let's have one enum that contains a string (reason for exit).

crates/corro-agent/src/plumtree/mod.rs
```diff
@@ -414,49 +454,24 @@ pub async fn plumtree_loop(
...
     loop {
-        let branch = tokio::select! {
-            biased;
-            _ = &mut tripwire => {
+        let branch = next_plumtree_loop_branch(
+            &mut tripwire,
+            &mut rx_plumtree,
+            &mut rx_plumtree_updates,
+            &mut plumtree_timer_rx,
+            &mut ihave_tick_interval,
```
somtochiama, line 463: let's also set the missed tick behaviour for ihave_tick_interval to skip

crates/corro-agent/src/plumtree/mod.rs
ZmoleCristian: Done in 4f21b32e: one Exit(reason) arm, handle_shutdown on every exit.

crates/corro-agent/src/plumtree/mod.rs
ZmoleCristian: Done in 4f21b32e

crates/corro-agent/src/plumtree/mod.rs
ZmoleCristian: Before swapping I tried to check what the maintenance tick reconciles, and I may be missing a path, so correct me if so. From what I can see it covers a first join and ring refresh, but not a departure, and not a rejoin on the same ring. If that is right, then under sustained input an update is not delayed a little, it is not polled, and past 256 queued it gets dropped.

From my measurements the swap buys about 0.4µs per iteration, and the current order costs at worst a burst of 256 updates, under a millisecond, before input resumes.

So I would propose one of:

1. keep the current order: updates above input, both below the timers
2. your order plus a bound: after N consecutive inputs, one try_recv on updates
3. your order plus a real reconciliation in the tick, plum-foca and its own tests, so probably a separate PR

Happy to go whichever way you prefer. The other two comments are in 4f21b32e.

<details>
<summary>What I read on departures and rejoins</summary>

As I read it, `update_peer_topology` only iterates the members snapshot and commits what it has not seen (`plum-foca/src/lib.rs:1024-1094`). I could not find anything that removes a known peer absent from the snapshot, so a `MemberDown` that is not applied would leave the peer in `known_peers`, in eager or lazy, and possibly ring locked, until restart.

`peer_down` seems to clear everything except `peer_topology` (`lib.rs:1102-1111`), so a peer rejoining on the same ring would hit the "already seen" path and not be re-added by the tick. A bucket change needs 8 confirmations (`lib.rs:427`), eight minutes at 60s.

If so, the best-effort comment at `handlers.rs:500` holds for joins and rings, but not for departures or rejoins.
</details>

<details>
<summary>What I measured on the production selector</summary>

Input producers use `send().await` (`agent/uni.rs:159-161`), so while they outpace the loop the channel stays non-empty.

Requested order, 4 producers, 2µs of work per input: a queued `MemberDown` was not serviced within 3s, several hundred thousand inputs went first, and it was picked up only after the whole 512 backlog drained once the producers stopped. Past 256 queued, `try_send` at `handlers.rs:501-503` drops updates.

Current order, same load: the update was serviced in tens of microseconds with nothing ahead of it. Updates are one per foca transition, so the worst case I could construct is 256 queued, drained in under a millisecond. With the updates channel empty the current order pays one extra poll per iteration, about 0.4µs, which I could only measure with a zero-work consumer.

The probes and the state machine tests behind these numbers are in a scratch branch if you want to look.
</details>

somtochiama: > but not a departure, and not a rejoin on the same ring.

Okay that's a bug. It should definitely take care of departures and rejoins too. `peer_topology` is supposed to be cache of all rtt values so we can place a peer well after they rejoin. We could check `known_peers` instead

somtochiama: It should be fine to keep the order in this pull request. Member Up/ Member Down takes v. little time (mostly updates internal state and then returns). During normal operations, events shouldn't happen often.
maintenance tick should still  reconcile plumtree peers to members tho.

ZmoleCristian: ✅ Keeping the order then. 

Thanks for the pointer on peer_topology, that makes the fix clear: reconcile against known_peers in the tick, drop known peers absent from the snapshot, leave peer_topology as the RTT cache. 
I will open that as a separate issue and PR.

PR #553, opened 2026-08-31
Run change-handler maintenance independently of batch timeout

### LINKED ISSUE #552: Seen-cache maintenance stops while the change handler has a backlog
## Description

The inbound change handler trims its recent-`seen` deduplication cache only from the batch wait-timeout branch. The burst scheduler intentionally disables that timeout while a batch is running and immediately starts another batch when enough work remains queued.

During sustained catch-up, the timeout can therefore stay disabled indefinitely.
Distinct actor/version keys continue to enter `seen`, but the cache-trim threshold
derived from `processing_queue_len` is not enforced by maintenance. The queue gauges
and `MetricsTracker` queue-size value in the same branch also stop updating.

There is a correctness consequence when processing fails. A change enters `seen` before its SQLite batch runs. If that batch returns an error, the version is not recorded in the bookie, but its cache entry remains. A retry received while backlog continues is rejected as a recent duplicate.

## Relevant path

In `crates/corro-agent/src/agent/handlers.rs`:

- `handle_new_change` sets `max_wait = None` when it spawns a threshold-sized batch.
- `handle_task_completion` immediately spawns the next batch when the backlog is above the threshold and does not recreate `max_wait`.
- `handle_changes` performs metrics reporting and `seen.drain(...)` only when `max_wait` fires.
- The duplicate-cache lookup runs before the bookie lookup and before processing.

## Reproduction

A deterministic test can exercise the scheduling behavior without exposing the cache:

1. Set `processing_queue_len = 4`, batch size to 2, the threshold ratio to `1.0`,
   and `apply_queue_timeout = 20 ms`.
2. Hold a separate SQLite `BEGIN IMMEDIATE` transaction and configure the pooled writer with a 100 ms busy timeout.
3. Start the real `handle_changes` task and send ten distinct versions from another actor.
4. Keep SQLite locked for 150 ms. The first two versions fail with
   `database is locked`, while the later versions keep the handler backlogged and grow
   `seen` beyond four keys.
5. Release SQLite and wait for the retained tail of the queue to apply.
6. Resend the failed first version.

Actual result: the resend is filtered by `seen` and does not reach the bookie before the test deadline.

Expected result: cache maintenance continues independently of batching state, evicts the old entry, and allows the unbooked version to be retried.

## Proposed fix

Run queue/cache maintenance on its own periodic interval rather than on the optional batch wait sleep. Put that interval ahead of hot input branches in the biased `select!` so it cannot be starved. Keep the existing retention policy and cadence.

## Impact

- The `seen` cache is not maintained at its configured scale during sustained backlog.
- Failed versions can be suppressed as duplicates for the duration of that backlog.
- Change-queue telemetry can remain stale during the period when it is most needed.

This report does not attach a measured RSS estimate; memory usage depends on workload shape. Exceeding the configured cache scale and suppressing the retry were reproduced locally through the real handler. Continued growth during a sustained backlog follows from the verified absence of any active maintenance path.

ZmoleCristian:
## Summary

- add a dedicated maintenance interval to `handle_changes`;
- move seen-cache trimming and queue metric updates to that interval; and
- cover failed-version retry under sustained backlog with a real-handler regression test.

## Why

The burst scheduler disables its optional `max_wait` sleep while processing full batches. Cache trimming and metric updates were attached to that same sleep, so both stopped for as long as enough work remained queued.

That can suppress recovery from a failed batch: the version is absent from the bookie but still present in the recent-`seen` cache, and subsequent copies are discarded as duplicates.

## Implementation

`handle_changes` now owns a Tokio interval at the configured `apply_queue_timeout` cadence. Maintenance is the first branch of the biased `select!`, which ensures a ready tick wins over completed jobs and a hot inbound channel. Missed ticks use `Skip`, and the maintenance duration has a 1 ms floor because Tokio rejects zero-duration intervals.

The existing cache threshold and retention calculation are unchanged. The optional `max_wait` sleep remains responsible only for flushing sub-threshold batches.

## Coordination

This may have a mechanical conflict with #527, which refactors `handle_changes`. If #527 lands first, I am happy to port this fix to the new actor layout.

## Test

`test_seen_cache_trim_runs_while_backlogged`:

- starts the real handler and SQLite processing path;
- forces the first two versions to fail with a bounded SQLite busy lock;
- sends enough later versions to exceed the seen-cache cap while work remains backlogged;
- releases SQLite; and
- verifies that a second copy of the failed version is accepted and applied.

The test fails on `936de111` with a retry deadline and passes with this change.

## Validation

- targeted regression: passed, including five consecutive runs;
- `cargo test -p corro-agent`: 51 passed, 2 ignored;
- `cargo clippy --all-targets -- -D warnings`: passed;
- `cargo fmt --all -- --check`: passed;
- `RUSTFLAGS='--cfg tokio_unstable' cargo nextest run --profile ci --workspace --target x86_64-unknown-linux-gnu`: 154 passed, 10 skipped.
- final verification on the committed tree: 154/154 executed CI-profile tests
  passed, with 10 configured skips; clippy, formatting, and diff checks passed.

## Performance and risk

This restores maintenance independence that existed before burst batching and does not
alter batch sizing, queue shedding, deduplication keys, or cache retention amounts. It
does add a periodic wake-up alongside the batching timeout. At the default 10 ms
cadence, the branch performs three gauge updates, one queue-size observation, and a
constant-time cache-size check; an actual trim is linear in entries removed. Missed
ticks are skipped. No throughput benchmark was run, so this draft does not characterize
the runtime cost as zero or negligible.

Fixes #552

### DIFF
--- a/crates/corro-agent/src/agent/handlers.rs
+++ b/crates/corro-agent/src/agent/handlers.rs
@@ -991,6 +991,8 @@ pub async fn handle_changes(
     mut tripwire: Tripwire,
 ) {
     let max_queue_len: usize = agent.config().perf.processing_queue_len;
+    let maintenance_duration =
+        Duration::from_millis(agent.config().perf.apply_queue_timeout.max(1) as u64);
 
     // Initialize batch processor state
     let mut state = HandleChangesState::new(
@@ -1014,10 +1016,27 @@ pub async fn handle_changes(
         0
     };
     let mut seen: IndexMap<_, RangeInclusiveSet<CrsqlSeq>> = IndexMap::new();
+    let mut maintenance_interval = tokio::time::interval(maintenance_duration);
+    maintenance_interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
     loop {
         let (change, src, original_bcast) = tokio::select! {
             biased;
+
+            // Queue metrics and cache maintenance must keep running while a
+            // processing task and its backlog keep the batching timeout disabled.
+            _ = maintenance_interval.tick() => {
+                gauge!("corro.agent.changes.in_queue").set(state.buf_cost as f64);
+                gauge!("corro.agent.changesets.in_queue").set(state.queue.len() as f64);
+                gauge!("corro.agent.changes.processing.jobs").set(if state.processing_task.is_some() { 1.0 } else { 0.0 });
+                metrics_tracker.observe_queue_size(state.queue.len() as u64);
+
+                if seen.len() > max_seen_cache_len {
+                    seen.drain(..seen.len() - keep_seen_cache_size);
+                }
+
+                continue;
+            },
 
             // Processing task finished
             res = async { state.processing_task.as_mut().unwrap().await }, if state.processing_task.is_some() => {
@@ -1033,18 +1052,7 @@ pub async fn handle_changes(
 
             // Timeout fires
             _ = async { state.max_wait.as_mut().unwrap().await }, if state.max_wait.is_some() => {
-                // Emit metrics
-                gauge!("corro.agent.changes.in_queue").set(state.buf_cost as f64);
-                gauge!("corro.agent.changesets.in_queue").set(state.queue.len() as f64);
-                gauge!("corro.agent.changes.processing.jobs").set(if state.processing_task.is_some() { 1.0 } else { 0.0 });
-                metrics_tracker.observe_queue_size(state.queue.len() as u64);
-
                 state.handle_timeout(&agent, &bookie);
-
-                // Cleanup seen cache
-                if seen.len() > max_seen_cache_len {
-                    seen.drain(..seen.len() - keep_seen_cache_size);
-                }
 
                 continue;
             },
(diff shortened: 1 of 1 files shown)


gorbak25 (dismissed):
lgtm

gorbak25:
@ZmoleCristian CI fails on macos, can u investigate?

ZmoleCristian:
> @ZmoleCristian CI fails on macos, can u investigate?

On it right now!

ZmoleCristian:
> @ZmoleCristian CI fails on macos, can u investigate?

@gorbak25 - Solved ✅
The problem: busy_timeout does not mean the same thing on macOS. BEGIN
IMMEDIATE with a 100ms timeout gave up at 197, 204, 227 and 251ms there,
against exactly 100ms on Linux.

My mistake: the test held the lock 150ms against that 100ms timeout, so it
needed the busy handler to be punctual.

Fixed by taking SQLite out of the setup. The test holds the pool's single
write permit, like the corro-pg tests do, so the backlog is structural
instead of timed. Handler change untouched, only the test. macOS 5/5 and the
full suite green, Linux 25/25, and it still fails against the unpatched
handler.

And I wired a macOS job on my fork so this stops reaching you.

PR #557, opened 2026-08-31
fix: fail `consul sync` without configuration

### LINKED ISSUE #556: `corrosion consul sync` exits successfully without Consul config
Was trying the CLI commands with `config.example.toml`, which does not have a `[consul]` block:

```rust
$ corrosion --config config.example.toml consul sync
ERROR corrosion: missing `consul` block in corrosion config
$ echo $?
0
```

The command cannot start syncing, but its successful exit status tells scripts and supervisors that it did. Running the agent without Consul configured is valid, but explicitly running `consul sync` without its required configuration should fail.

Digged thru the issues and PRs, this is similar to [#481](https://github.com/superfly/corrosion/pull/481), which made `corrosion reload` return an error instead of only logging it.

Imo, when `consul sync` is run without a `[consul]` config block, it should print the configuration error and exit with a non zero status, but currently it prints the error and exits successfully.

wthrajat:
Attempts to fix https://github.com/superfly/corrosion/issues/556

Now the `corrosion consul sync` returns the missing configuration error, so the CLI exits with a non zero status instead of reporting success

PS: the regression test runs the command with `config.example.toml` and checks both the exit status and error message.

### DIFF
--- a/crates/corrosion/src/main.rs
+++ b/crates/corrosion/src/main.rs
@@ -395,9 +395,7 @@ async fn process_cli(cli: Cli) -> eyre::Result<()> {
                 Some(consul) => {
                     command::consul::sync::run(consul, cli.api_addr()?, cli.db_path()?).await?
                 }
-                None => {
-                    error!("missing `consul` block in corrosion config");
-                }
+                None => eyre::bail!("missing `consul` block in corrosion config"),
             },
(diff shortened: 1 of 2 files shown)


gorbak25 (approved):
lgtm

PR #570, opened 2026-09-12
fix(corro-types): re-key by_addr when a member's gossip address changes

### LINKED ISSUE #569: Members::add_member does not re-key by_addr when a member's gossip address changes, freezing ring recalculation
## Summary

`Members::add_member` (corro-types/src/members.rs) handles an update for an existing `ActorId` whose gossip address changed (same actor, newer `ts`) by overwriting `member.addr`, but never updates the `by_addr` reverse index:

- the old address still maps to the actor id,
- the new address has no entry,
- `recalculate_rings()` looks members up through `by_addr`, so ring recalculation silently no-ops for the moved member.

The same call path also leaves the `rtts` window keyed by the old address, so future RTT samples for the new address start an empty window and existing samples are stranded under a stale key.

## Impact

When a SWIM identity renews with a new address (e.g. a Fly.io redeploy), the member's ring freezes: no ring change can ever be computed from the new address, and the stale `by_addr` entry for the old address leaks until the member is removed. If another member later binds the old address, `add_member`'s `member_id` mismatch branch removes the index entry at the *incoming* address rather than the stored one, which can clobber the other member's `by_addr` entry.

## Reproduction sketch

1. `Members::add_member` with actor A at addr1 (member_id matching).
2. `add_member` with the same actor A, addr2, newer ts: the result is `Updated` and `states[A].addr == addr2`, but `by_addr` still contains `addr1 -> A` and has no `addr2` entry.
3. `add_rtt(addr2, ...)` never updates A's ring because `recalculate_rings(addr2)` finds nothing in `by_addr`.

## Suggested fix

In the `Ordering::Greater` branch, when the address changed: remove the old `by_addr` entry (guarded so it only removes when it maps to this actor id), insert the new one, move the `rtts` window to the new address, and recalculate rings. In the `member_id` mismatch branch, remove the index entry under the member's stored address rather than the incoming actor's address.

I have a patch with tests ready and will open a PR referencing this issue.

Tyagiquamar:
Fixes #569

## Problem

`Members::add_member`'s update branch (`Ordering::Greater`) overwrote `MemberState.addr` without touching the `by_addr` reverse index. After a SWIM identity renewed on a new address (e.g. a Fly.io redeploy):

- the old address still mapped to the actor id in `by_addr`,
- the new address had no entry,
- `recalculate_rings()` resolves members through `by_addr`, so it silently no-oped for the moved member and its ring froze,
- the `rtts` window stayed keyed by the old address, stranding existing samples.

The `member_id` mismatch branch had a related defect: it removed the `by_addr` entry at the *incoming* actor's address instead of the member's stored address, leaking the removed member's entry and clobbering another member's entry when they shared an address.

## Fix

- In the update branch, when the address changes: remove the old `by_addr` entry (guarded so it only removes when it maps to this actor id), insert the new one, move the `rtts` window to the new address, and recalculate rings for it.
- In the mismatch branch, remove the `by_addr` entry under the member's stored address, with the same guard.

## Tests

Three new tests in `members.rs`, all failing before the fix:

- `update_to_new_address_rekeys_by_addr`
- `rtt_samples_follow_the_member_to_its_new_address`
- `mismatched_member_id_removes_the_index_entry_for_the_stored_address`

Verified in a `rust:1.89` Linux container matching CI: `cargo test -p corro-types --lib` 60/60 pass, `cargo clippy -p corro-types --all-targets -- -D warnings` clean, `cargo fmt --check` clean, `cargo check --workspace` green.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

